PR #2, opened 2016-04-13
Cleanup formatting and minor code changes

ViToni:
pfSense Ticket #5783
https://redmine.pfsense.org/issues/5783

see:
https://github.com/pfsense/FreeBSD-ports/commit/28cdacad47051a02e7866093634de3326242d21e

### DIFF
--- a/src/igmpproxy.c
+++ b/src/igmpproxy.c
@@ -215,7 +215,7 @@ int igmpProxyInit(void) {
         }
 
 	if(0 == upsvifcount) {
-            my_log(LOG_ERR, 0, "There unless be 1 Vif's as upstream.");
+            my_log(LOG_ERR, 0, "There must be at least 1 Vif as upstream.");
         }
     }  
     
--- a/src/igmpproxy.c
+++ b/src/igmpproxy.c
@@ -134,7 +134,7 @@ int main( int ArgCn, char *ArgVc[] ) {
 	    // Detach daemon from terminal
 	    if ( close( 0 ) < 0 || close( 1 ) < 0 || close( 2 ) < 0
 		 || open( "/dev/null", 0 ) != 0 || dup2( 0, 1 ) < 0 || dup2( 0, 2 ) < 0
-		 || setpgrp() < 0
+		 || setpgid( 0, 0 ) < 0
 	       ) {
 		my_log( LOG_ERR, errno, "failed to detach daemon" );
 	    }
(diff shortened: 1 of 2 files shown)


pali:
Hi! Please add proper description into commit message and do not add external links... Also maybe you should squash them into one commit.

ViToni:
Could you please check again.

src/ifvc.c
```diff
@@ -268,6 +270,20 @@ void buildIfVc(void) {
             mask = ((struct sockaddr_in *)&IfReq.ifr_addr)->sin_addr.s_addr;
             subnet = addr & mask;
 
+            dp = getIfByName(IfPt->ifr_name, 1);
+            if (dp != NULL && dp->allowednets != NULL) {
+                allowednet = (struct SubnetList *)malloc(sizeof(struct SubnetList));
+                if (allowednet == NULL) my_log(LOG_ERR, 0, "Out of memory !");
```
pali: this is broken code

pali: Please fix this part and also fix your email address in your git commit.

ViToni: Fixed the formatting related to this change.

The email adress is the same as used by changes done inside the GitHub UI. Which one is prefered for consistency (GitHub email or private email)?

pali: Null pointer dereference is still not fixed. Also rebase patch on top of branch.

Email address must be valid. Does not matter if private or public... Just valid address on which people can reach you (e.g. in case somebody would need to update your patch).

ViToni:
Rewrote all email addresses so that only one email is used for changes by my 
Removed the code in question. The puill request consists now mainly of cleanups, chnage of incompatible code und removal of code duplicates.
Especially the code formatting should ease further patches.

src/ifvc.c
ViToni: Maybe you could explain the problem in the above code to me, because my C is quite rusty and I clearly fail to see it.

pali: Null pointer dereference:

allowednet = ...
if (allowednet == NULL) my_log(...);
allowednet->...

pali:
"This branch has conflicts that must be resolved"

And please do different things in different commits/pull requests.
Also include description into commit message. People should understand it without need to open external links...

src/ifvc.c
ViToni: I believe code above avoids NULL pointer dereference because logging an error via 

> `if (allowednet == NULL) my_log(LOG_ERR, 0, "Out of memory !");`

leads to 'src/syslog.c' and there it exits after logging:

> `if( Severity <= LOG_ERR )
>         exit( -1 );
> `

PR #6, opened 2016-05-19
Use standard unsigned integer types

matthew-l-weber:
Use the standard unsigned integer types instead of the non-standard
u_char, u_short and u_int.
(This fixes builds using the musl libc)

Based on:
https://github.com/alpinelinux/aports/blob/master/main/igmpproxy/0001-Use-standard-unsigned-integer-types.patch

Signed-off-by: Bernd Kuhls bernd.kuhls@t-online.de
Signed-off-by: Matt Weber matthew.weber@rockwellcollins.com

### DIFF
--- a/src/igmp.c
+++ b/src/igmp.c
@@ -79,7 +79,7 @@ void initIgmp() {
 /**
 *   Finds the textual name of the supplied IGMP request.
 */
-char *igmpPacketKind(u_int type, u_int code) {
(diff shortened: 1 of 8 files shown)


pali:
Similar patch is already in tree, see commit 33a69727c936e81aff312d1f92ab5681a77b095d. And it is in conflict with yours.

matthew-l-weber:
Opps, just realized you use a next branch.  Ignore this and I'll close the request.

PR #8, opened 2016-05-19
uclinux support

matthew-l-weber:
Added support for building against a uclinux target.

Signed-off-by: Matthew Weber matthew.weber@rockwellcollins.com


pali:
Again, there is merge conflict with commit 0e3af272c87d77cc81740436fcf6bc69e0c3c215. And it is exactly same as yours. I would suggest you to rebase your patches on top of branch. Using git rebase will skip commits which have same differences as some existing commit...

PR #14, opened 2016-10-24
Added extended logging and code cleanup

ViToni:


src/rttable.c
```diff
@@ -712,40 +702,40 @@ int internUpdateKernelRoute(struct RouteTable *route, int activate) {
...
+                croute->ageValue, st,
+                croute->vifBits);
+
+            croute = croute->nextroute; 
+
+            rcount++;
+        } while ( croute != NULL );
+    }
+
+    my_log(LOG_DEBUG, 0, "-----------------------------------------------------");
```
pali, line 788: Hi! I think that this commit 3fb62e50fc9b953f73de8eaaaa12ed464c215327 does not clean up code and make it consistent. E.g. in this line you do not use spaces around brackets...

src/udpsock.c
```diff
@@ -49,17 +49,19 @@ int openUdpSocket( uint32_t PeerInAdr, uint16_t PeerPort ) {
...
+    }
+
     memset( &SockAdr, 0, sizeof( SockAdr ) );
     SockAdr.sin_family      = AF_INET;
     SockAdr.sin_port        = htons(PeerPort);
     SockAdr.sin_addr.s_addr = htonl(PeerInAdr);
-    
-    if( bind( Sock, (struct sockaddr *)&SockAdr, sizeof( SockAdr ) ) )
+
+    if( bind( Sock, (struct sockaddr *)&SockAdr, sizeof( SockAdr ) ) ) {
```
pali, line 61: ... and on this line you are using spaces around brackets.

src/mroute-api.c
```diff
@@ -65,9 +65,16 @@ static struct VifDesc {
 **          - the errno value for non-fatal failure condition
 */
 int enableMRouter(void) {
-    int Va = 1;
+    int Va = 1, socket_type;
 
-    if ( (MRouterFD  = socket(AF_INET, SOCK_RAW, IPPROTO_IGMP)) < 0 ) {
+    socket_type = SOCK_RAW;
+
+#ifdef SOCK_CLOEXEC
```
pali, line 74: This is not a correct way... You can have SOCK_CLOEXEC defined, but compiled code will be running on older kernel without SOCK_CLOEXEC support. Correct way is to use runtime check, not only compile time.

pali:
I see that you are doing more cleanup and cosmetic changes in different commits on different places. I would suggest you to do whole coding style, comments and cosmetic changes in separated commits (without any function changes) and then do real work. As it is really hard to look at diff when there are mixing cosmetic changes and real code changes...

pali:
Now next branch was removed and code moved to master branch. Code in master is now GPLv2+ compatible, so please re-open pull request against master branch.

PR #19, opened 2017-05-05
Introduce default format for code base.

ViToni:
The first commit introduces a new goal: `make reformat` based on clang-reformat (at least version 3.9 should be used).
This target is optional.

The second commit contains all changes after the execution of this command.

In the future committers can normalize their code by simply using the make goal.

### DIFF
--- a/src/confread.c
+++ b/src/confread.c
@@ -45,22 +45,23 @@
 
 #include "igmpproxy.h"
 
-#define READ_BUFFER_SIZE    512     // Inputbuffer size...
+#define READ_BUFFER_SIZE 512 // Inputbuffer size...
 
 #ifndef MAX_TOKEN_LENGTH
-    #define MAX_TOKEN_LENGTH  30    // Default max token length
+#define MAX_TOKEN_LENGTH 30 // Default max token length
 #endif
 
-FILE            *confFilePtr;       // File handle pointer
-char            *iBuffer;           // Inputbuffer for reading...
-unsigned int    bufPtr;             // Buffer position pointer.
-unsigned int    readSize;           // Number of bytes in buffer after last read...
-char    cToken[MAX_TOKEN_LENGTH];   // Token buffer...
-short   validToken;
+FILE *confFilePtr;             // File handle pointer
+char *iBuffer;                 // Inputbuffer for reading...
+unsigned int bufPtr;           // Buffer position pointer.
+unsigned int readSize;         // Number of bytes in buffer after last read...
+char cToken[MAX_TOKEN_LENGTH]; // Token buffer...
+short validToken;
 
-/**
-*   Opens config file specified by filename.
-*/
+
+/*******************************************************************************
+ * Opens config file specified by filename.
+ ******************************************************************************/
 int openConfigFile(char *filename) {
 
     // Set the buffer to null initially...
@@ -70,14 +71,14 @@ int openConfigFile(char *filename) {
     confFilePtr = fopen(filename, "r");
 
     // On error, return false
-    if(confFilePtr == NULL) {
+    if (confFilePtr == NULL) {
         return 0;
     }
 
     // Allocate memory for inputbuffer...
-    iBuffer = (char*) malloc( sizeof(char) * READ_BUFFER_SIZE );
+    iBuffer = (char *)malloc(sizeof(char) * READ_BUFFER_SIZE);
 
-    if(iBuffer == NULL) {
+    if (iBuffer == NULL) {
         closeConfigFile();
         return 0;
     }
@@ -89,70 +90,74 @@ int openConfigFile(char *filename) {
     return 1;
 }
 
-/**
-*   Closes the currently open config file.
-*/
+/*******************************************************************************
+ * Closes the currently open config file.
+ ******************************************************************************/
 void closeConfigFile(void) {
     // Close the file.
-    if(confFilePtr!=NULL) {
+    if (confFilePtr != NULL) {
         fclose(confFilePtr);
     }
     // Free input buffer memory...
-    if(iBuffer != NULL) {
+    if (iBuffer != NULL) {
         free(iBuffer);
     }
 }
 
-/**
-*   Returns the next token from the configfile. The function
-*   return NULL if there are no more tokens in the file.
-*/
+
+/*******************************************************************************
+ * Returns the next token from the configfile.
+ * The function return NULL if there are no more tokens in the file.
+ ******************************************************************************/
 char *nextConfigToken(void) {
 
     validToken = 0;
 
     // If no file or buffer, return NULL
-    if(confFilePtr == NULL || iBuffer == NULL) {
+    if (confFilePtr == NULL || iBuffer == NULL) {
         return NULL;
     }
 
     {
-        unsigned int tokenPtr       = 0;
-        unsigned short finished     = 0;
+        unsigned int tokenPtr = 0;
+        unsigned short finished = 0;
         unsigned short commentFound = 0;
 
         // Outer buffer fill loop...
-        while ( !finished ) {
-            // If readpointer is at the end of the buffer, we should read next chunk...
-            if(bufPtr == readSize) {
+        while (!finished) {
+            /*
+             *  If readpointer is at the end of the buffer, we should read next
+             *  chunk...
+             */
+            if (bufPtr == readSize) {
                 // Fill up the buffer...
-                readSize = fread (iBuffer, sizeof(char), READ_BUFFER_SIZE, confFilePtr);
+                readSize = fread(iBuffer, sizeof(char), READ_BUFFER_SIZE, confFilePtr);
                 bufPtr = 0;
 
                 // If the readsize is 0, we should just return...
-                if(readSize == 0) {
+                if (readSize == 0) {
                     return NULL;
                 }
             }
 
             // Inner char loop...
-            while ( bufPtr < readSize && !finished ) {
+            while (bufPtr < readSize && !finished) {
 
-                //printf("Char %s", iBuffer[bufPtr]);
+                // printf("Char %s", iBuffer[bufPtr]);
 
                 // Break loop on \0
-                if(iBuffer[bufPtr] == '\0') {
+                if (iBuffer[bufPtr] == '\0') {
                     break;
                 }
 
-                if( commentFound ) {
-                    if( iBuffer[bufPtr] == '\n' ) {
+                if (commentFound) {
+                    if (iBuffer[bufPtr] == '\n') {
                         commentFound = 0;
                     }
                 } else {
 
                     // Check current char...
-                    switch(iBuffer[bufPtr]) {
+                    switch (iBuffer[bufPtr]) {
                     case '#':
                         // Found a comment start...
                         commentFound = 1;
@@ -163,8 +168,8 @@ char *nextConfigToken(void) {
                     case '\t':
                     case ' ':
                         // Newline, CR, Tab and space are end of token, or ignored.
-                        if(tokenPtr > 0) {
-                            cToken[tokenPtr] = '\0';    // EOL
+                        if (tokenPtr > 0) {
+                            cToken[tokenPtr] = '\0'; // EOL
                             finished = 1;
                         }
                         break;
@@ -177,7 +182,7 @@ char *nextConfigToken(void) {
                 }
 
                 // Check end of token buffer !!!
-                if(tokenPtr == MAX_TOKEN_LENGTH - 1) {
+                if (tokenPtr == MAX_TOKEN_LENGTH - 1) {
                     // Prevent buffer overrun...
                     cToken[tokenPtr] = '\0';
                     finished = 1;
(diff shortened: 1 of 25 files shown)


pali:
It is a good idea to have just one consistent coding style in whole project, but I dislike your proposed code style where are too many spaces around parenthesis. E.g. spaces around void in
`static void debugQueue( void );`

ViToni:
Actually I don't mind which formatting style is to be used.
For me it's only important that there is one and that formatting is consistent all over the code base.

There are even some defaults styles (`LLVM, Google, Chromium, Mozilla, WebKit`) which could used:
[http://releases.llvm.org/3.9.0/tools/clang/docs/ClangFormatStyleOptions.html]()

Just choose a style and we can stick to it.

pali:
Try something like this?
```
BasedOnStyle: LLVM
Language: Cpp
IndentWidth: 4
ColumnLimit: 120
AllowShortFunctionsOnASingleLine: Empty
AlwaysBreakBeforeMultilineStrings: true
```

ViToni:
AFAIK `AllowShortFunctionsOnASingleLine` should be either `true` or `false`.

I would suggest having `UseTab: Never` as one of the mandatory options as the mixture of tabs and whitespaces really messes diffs up.

Having also `MaxEmptyLinesToKeep: 1` or `MaxEmptyLinesToKeep: 2` might be useful, too.

ViToni:
BTW I already pushed another version of reformatted source code settings (based on the initial format settings) when writting https://github.com/pali/igmpproxy/pull/19#issuecomment-299673664 , so you might want to check, if you like it more.

pali:
> AFAIK `AllowShortFunctionsOnASingleLine` should be either `true` or `false`

According to http://releases.llvm.org/3.9.0/tools/clang/docs/ClangFormatStyleOptions.html possible values are: `None`, `Empty`, `Inline`, `All`

> I would suggest having `UseTab: Never`

Already part of `BasedOnStyle: LLVM`

> Having also `MaxEmptyLinesToKeep: 1`

Also part of `BasedOnStyle: LLVM`

> or `MaxEmptyLinesToKeep: 2` might be useful, too.

`BasedOnStyle: LLVM` set it to 1, but 2 seems to be better value.

ViToni:
>> AFAIK `AllowShortFunctionsOnASingleLine` should be either `true` or `false`

> According to http://releases.llvm.org/3.9.0/tools/clang/docs/ClangFormatStyleOptions.html possible values are: `None`, `Empty`, `Inline`, `All`

You are right. (saw the `boolean` option here [https://clangformat.com/]())

>> I would suggest having `UseTab: Never`

> Already part of `BasedOnStyle: LLVM`

It was more meant to be explicit.

I update the PR using these settings:

    BasedOnStyle: LLVM
    Language: Cpp
    
    IndentWidth: 4
    ContinuationIndentWidth: 8
    
    ColumnLimit: 120
    
    MaxEmptyLinesToKeep: 2
    
    AllowShortFunctionsOnASingleLine: Empty
    AlwaysBreakBeforeMultilineStrings: true

src/ifvc.c
```diff
@@ -170,89 +173,87 @@ void rebuildIfVc () {
...
     }
 
-    close( Sock );
+    close(Sock);
 }
 
 /*
-** Builds up a vector with the interface of the machine. Calls to the other functions of
+** Builds up a vector with the interface of the machine. Calls to the other
+*functions of
```
pali: This is badly wrapped comment. Such changes are really not acceptable.

ViToni:
Changed the mentioned part manually so that future calls of `make reformat` should work there nicely.

src/confread.c
```diff
@@ -112,47 +112,48 @@ char *nextConfigToken(void) {
...
+        unsigned short finished = 0;
         unsigned short commentFound = 0;
 
         // Outer buffer fill loop...
-        while ( !finished ) {
-            // If readpointer is at the end of the buffer, we should read next chunk...
-            if(bufPtr == readSize) {
+        while (!finished) {
+            // If readpointer is at the end of the buffer, we should read next
+            // chunk...
```
pali: same there

src/igmp.c
```diff
@@ -136,39 +139,42 @@ void acceptIgmp(int recvlen) {
...
-            {
-                if(-1 != upStreamIfIdx[i])
-                {
-                    // Check if the source address matches a valid address on upstream vif.
-                    checkVIF = getIfByIx( upStreamIfIdx[i] );
-                    if(checkVIF == 0) {
+            for (i = 0; i < MAX_UPS_VIFS; i++) {
+                if (-1 != upStreamIfIdx[i]) {
+                    // Check if the source address matches a valid address on upstream
+                    // vif.
```
pali: same here

src/igmp.c
```diff
@@ -136,39 +139,42 @@ void acceptIgmp(int recvlen) {
...
+                                   "not in any valid net for upstream "
+                                   "VIF[%d].",
+                                   inetFmt(src, s1), inetFmt(dst, s2), i);
                         }
                     } else {
                         // Activate the route.
                         int vifindex = checkVIF->index;
-                        my_log(LOG_DEBUG, 0, "Route activate request from %s to %s on VIF[%d]",
-                            inetFmt(src,s1), inetFmt(dst,s2), vifindex);
+                        my_log(LOG_DEBUG, 0, "Route activate request from %s to %s on VIF[%d]", inetFmt(src, s1),
```
pali, line 178: rather let `inetFmt(src, s1)` at second line after wrap

src/mcgroup.c
```diff
@@ -71,17 +65,19 @@ static int joinleave( int Cmd, int UdpSock, struct IfDesc *IfDp, uint32_t mcasta
 *   The join is bound to the UDP socket 'UdpSock', so if this socket is
 *   closed the membership is dropped.
 *
-*   @return 0 if the function succeeds, 1 if parameters are wrong or the join fails
+*   @return 0 if the function succeeds, 1 if parameters are wrong or the join
+* fails
```
pali: same here

src/igmp.c
ViToni: This is the result of automatic reformatting.

It can be changed either by having other formatting options (which would still look differently) or  
it could be solved by using clang directives such as

    int formatted_code;
    // clang-format off
        void    unformatted_code   ;
    // clang-format on
    void formatted_code_again;

I would prefer rather not introduce clang specific stuff into the code.

pali: So what does it mean? It means that automatic reformatting is not a perfect solution and has problems.

Therefore we should use it only as help tool, not enforcement, like here.

src/igmp.c
```diff
@@ -136,39 +139,44 @@ void acceptIgmp(int recvlen) {
...
                         my_log(LOG_ERR, 0, "Upstream VIF was null.");
                         return;
-                    }
-                    else if(src == checkVIF->InAdr.s_addr) {
-                        my_log(LOG_NOTICE, 0, "Route activation request from %s for %s is from myself. Ignoring.",
-                            inetFmt(src, s1), inetFmt(dst, s2));
+                    } else if (src == checkVIF->InAdr.s_addr) {
+                        my_log(LOG_NOTICE, 0,
+                               "Route activation request from %s for %s is "
+                               "from myself. Ignoring.",
```
pali, line 158: Please do not split string into more multiline parts.

src/igmp.c
```diff
@@ -136,39 +139,44 @@ void acceptIgmp(int recvlen) {
...
-                    }
-                    else if(!isAdressValidForIf(checkVIF, src)) {
+                    } else if (!isAdressValidForIf(checkVIF, src)) {
                         struct IfDesc *downVIF = getIfByAddress(src);
                         if (downVIF && downVIF->state & IF_STATE_DOWNSTREAM) {
-                            my_log(LOG_NOTICE, 0, "The source address %s for group %s is from downstream VIF[%d]. Ignoring.",
-                                inetFmt(src, s1), inetFmt(dst, s2), i);
+                            my_log(LOG_NOTICE, 0,
+                                   "The source address %s for group %s is "
+                                   "from downstream VIF[%d]. Ignoring.",
```
pali, line 166: same here (and on other places)

src/igmpproxy.c
```diff
@@ -40,37 +40,35 @@
...
-"   -v   Be verbose. Give twice to see even debug messages.\n"
-"\n"
-PACKAGE_STRING "\n"
-;
+        "Usage: igmpproxy [-h] [-d] [-v [-v]] <configfile>\n"
+        "\n"
+        "   -h   Display this help screen\n"
+        "   -d   Run in debug mode. Output all messages on stderr\n"
+        "   -v   Be verbose. Give twice to see even debug messages.\n"
+        "\n" PACKAGE_STRING "\n";
```
pali, line 48: and here should be newline, after each "\n" to visually identify end of lines.

src/igmpproxy.c
```diff
@@ -40,37 +40,35 @@
...
-void    igmpProxyCleanUp(void);
-void    igmpProxyRun(void);
+int igmpProxyInit(void);
+void igmpProxyCleanUp(void);
+void igmpProxyRun(void);
 
 // Global vars...
 static int sighandled = 0;
-#define GOT_SIGINT  0x01
-#define GOT_SIGHUP  0x02
```
pali, line 61: let that space here, to have visually aligned those #defines

src/igmpproxy.c
```diff
@@ -104,40 +102,39 @@ int main( int ArgCn, char *ArgVc[] ) {
...
+                exit(0);
 
             // Detach daemon from terminal
-            if ( close( 0 ) < 0 || close( 1 ) < 0 || close( 2 ) < 0
-                || open( "/dev/null", 0 ) != 0 || dup2( 0, 1 ) < 0 || dup2( 0, 2 ) < 0
-                || setpgid( 0, 0 ) < 0
-            ) {
-                my_log( LOG_ERR, errno, "failed to detach daemon" );
+            if (close(0) < 0 || close(1) < 0 || close(2) < 0 || open("/dev/null", 0) != 0 || dup2(0, 1) < 0 ||
+                dup2(0, 2) < 0 || setpgid(0, 0) < 0) {
```
pali, line 136: previous wrapping to 3 lines was better. close() calls on one line, open+dup() on second line and setpgid() on third.

src/igmpproxy.h
```diff
@@ -36,51 +36,50 @@
...
  * Limit on length of route data
  */
-#define MAX_IP_PACKET_LEN	576
-#define MIN_IP_HEADER_LEN	20
-#define MAX_IP_HEADER_LEN	60
-#define IP_HEADER_RAOPT_LEN	24
+#define MAX_IP_PACKET_LEN 576
+#define MIN_IP_HEADER_LEN 20
+#define MAX_IP_HEADER_LEN 60
+#define IP_HEADER_RAOPT_LEN 24
```
pali, line 69: And let those defines aligned as they were too

src/lib.c
```diff
@@ -48,12 +48,9 @@ char s4[19];
 ** returns: - pointer to 'St'
 **
 */
-char *fmtInAdr( char *St, struct in_addr InAdr ) {
-    sprintf( St, "%u.%u.%u.%u",
-             ((uint8_t *)&InAdr.s_addr)[ 0 ],
-             ((uint8_t *)&InAdr.s_addr)[ 1 ],
-             ((uint8_t *)&InAdr.s_addr)[ 2 ],
-             ((uint8_t *)&InAdr.s_addr)[ 3 ] );
```
pali, line 56: And line wrapping here was also better before reformatting

src/request.c
```diff
@@ -177,26 +170,27 @@ void sendGroupSpecificMemberQuery(void *argument) {
...
 
     /**
-     * FIXME: This loops through all interfaces the group is active on an sends queries.
-     *        It might be better to send only a query on the interface the leave was accepted on and remove only that interface from the route.
+     * FIXME:
+     * This loops through all interfaces the group is active on and sends
+     * queries.
+     *
+     * It might be better to send only a query on the interface the leave
+     * was accepted on and remove only that interface from the route.
```
pali, line 178: This is not good reformat. Second sentence is also part of FIXME, but after change is not visually anymore.

src/request.c
```diff
@@ -206,29 +200,24 @@ void sendGroupSpecificMemberQuery(void *argument) {
...
-                         IGMP_MEMBERSHIP_QUERY,
+                sendIgmp(Dp->InAdr.s_addr, allhosts_group, IGMP_MEMBERSHIP_QUERY,
                          conf->queryResponseInterval * IGMP_TIMER_SCALE, 0, 0);
 
-                my_log(LOG_DEBUG, 0,
-                    "Sent membership query from %s to %s. Delay: %d",
-                    inetFmt(Dp->InAdr.s_addr,s1),
-                    inetFmt(allhosts_group,s2),
-                    conf->queryResponseInterval);
+                my_log(LOG_DEBUG, 0, "Sent membership query from %s to %s. Delay: %d", inetFmt(Dp->InAdr.s_addr, s1),
```
pali, line 219: here also put inetFmt on second line

src/confread.c
```diff
@@ -45,22 +45,23 @@
...
+char cToken[MAX_TOKEN_LENGTH]; // Token buffer...
+short validToken;
 
-/**
-*   Opens config file specified by filename.
-*/
+
+/*******************************************************************************
+ * Opens config file specified by filename.
+ ******************************************************************************/
```
pali, line 64: This reformat (and on other places too) introduce inconsistency. As in other files is used just:
```
/**
*   ...
*   ...
*/
```

pali:
@ViToni seems that auto-reformat by robot/machine for everything does not work as expected. After automatic reformat it needs some manual work to fix some parts. So I do not think that "make reformat" target is a good idea.

But, this is really an improvement for the code... so are you going to fix needed parts manually and finish this work in this pull request?

PR #20, opened 2017-05-06
Improve logging

ViToni:
When logging to stderr timestamps are added to the output making debugging a lot easiser.

Additionally the short circuit log level evaluation avoid unnecessary function calls if these log calls wouldn't be shown because of the log level.
This feature might be especially useful for embedded devices.

This PR is based on PR #19.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,61 @@
+Language: Cpp
+
+IndentWidth: 4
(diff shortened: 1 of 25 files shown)


pali:
This PR depends on #19. Therefore in current state cannot be merged.

PR #25, opened 2018-01-11
Fix running as daemon

CyberShadow:
Commit 29eab814175d7754797c51ed6d4d1d7c91155cca introduced a new flag,
-n, which allows making igmpproxy not run as a daemon (fork and run in
the foreground). However, the boolean flag controlling this option was
always true, thus causing igmpproxy to never run as a daemon,
regardless whether the switch was specified or not.


pali:
You are right, thanks for spotting this.

pali:
Fixed now in new release 0.2.1: https://github.com/pali/igmpproxy/releases/tag/0.2.1

PR #32, opened 2018-05-03
fix timeout of age_callout_queue

scholarchen:
` timeout->tv_sec = (secs > 3) ? 3 : secs; // aimwang: set max timeout`
pselect return max timeout is 3 seconds not secs when Rt==0

### DIFF
--- a/src/igmpproxy.c
+++ b/src/igmpproxy.c
@@ -338,7 +338,7 @@ void igmpProxyRun(void) {
              * call gettimeofday.
              */
             if (Rt == 0) {
-                curtime.tv_sec = lasttime.tv_sec + secs;
+                curtime.tv_sec = lasttime.tv_sec + timeout->tv_sec;//timeout
                 curtime.tv_nsec = lasttime.tv_nsec;
                 Rt = -1; /* don't do this next time through the loop */
             } else {


src/igmpproxy.c
```diff
@@ -338,7 +338,7 @@ void igmpProxyRun(void) {
              * call gettimeofday.
              */
             if (Rt == 0) {
-                curtime.tv_sec = lasttime.tv_sec + secs;
+                curtime.tv_sec = lasttime.tv_sec + timeout->tv_sec;//timeout
```
pali, line 341: IIRC updating `timeout` value of pselect is Linux specific. So it would not work on other platforms correctly.

pali: Really manpage for pselect says:

On Linux, select() modifies timeout to reflect the amount of time not slept; **most other implementations do not do this**. (POSIX.1 permits either behavior.) This causes problems both when Linux code which reads timeout is ported to other operating systems, and when code is ported to Linux that reuses a struct timeval for multiple select()s in a loop without reinitializing it. **Consider timeout to be undefined after select() returns.**

So this patch is Linux-specific, but igmpproxy is used on more *BSD platforms.

pali (changes requested):

zocker007:
I have found that issue too and fixed it like in this PR. On FreeBSD, this patch works like expected.

I think this is the same as in #59 . When the solution there works like expected, I think we should probably use it over this because of portability.

pali:
I guess that instead of using undefined value `timeout->tv_sec` you can use `(secs > 3) ? 3 : secs` or call `clock_gettime` with `CLOCK_MONOTONIC`.

Issue #28, opened 2018-04-16
Igmpproxy is not starting if upstream interface is a pppoe interface


micharambou:
I'm running igmpproxy within OPNsense 18.1.6. 
Since 0.1_2,1 -> 0.2.1,1 update igmpproxy fails to start if upstream interface in igmpproxy.conf is a ~~vlan tagged~~ pppoe interface.

In my case uppstream Interface is pppoe on a vlan tagged interface. 

```
/usr/local/etc/igmpproxy.conf`

##------------------------------------------------------
## Enable Quickleave mode (Sends Leave instantly)
##------------------------------------------------------
quickleave
phyint pppoe0 upstream ratelimit 0 threshold 1
altnet 224.0.0.0/24
altnet 87.141.215.0/24

phyint igb0 downstream ratelimit 0 threshold 1
altnet 192.168.100.0/24
```

```
root@OPNsense:~ # /usr/local/sbin/igmpproxy -d -vv /usr/local/etc/igmpproxy.conf
Searching for config file at '/usr/local/etc/igmpproxy.conf'
Config: Quick leave mode enabled.
Config: Got a phyint token.
Config: IF: Config for interface pppoe0.
Config: IF: Got upstream token.
Config: IF: Got ratelimit token '0'.
Config: IF: Got threshold token '1'.
Config: IF: Got altnet token 224.0.0.0/24.
Config: IF: Altnet: Parsed altnet to 224.0.0/24.
Config: IF: Got altnet token 87.141.215.0/24.
Config: IF: Altnet: Parsed altnet to 87.141.215/24.
IF name : pppoe0
Next ptr : 0
Ratelimit : 0
Threshold : 1
State : 1
Allowednet ptr : 8182b000
Config: Got a phyint token.
Config: IF: Config for interface igb0.
Config: IF: Got downstream token.
Config: IF: Got ratelimit token '0'.
Config: IF: Got threshold token '1'.
Config: IF: Got altnet token 192.168.100.0/24.
Config: IF: Altnet: Parsed altnet to 192.168.100/24.
IF name : igb0
Next ptr : 0
Ratelimit : 0
Threshold : 1
State : 2
Allowednet ptr : 8182b020
Config: Got a phyint token.
Config: IF: Config for interface igb2.
Config: IF: Got disabled token.
IF name : igb2
Next ptr : 0
Ratelimit : 0
Threshold : 1
State : 0
Allowednet ptr : 0
[...]

PR #38, opened 2018-07-31
Added changes to permit pppoe interfaces

ljpaff:
Workaround for:
**Igmpproxy is not starting if upstream interface is a pppoe interface [#28](https://github.com/pali/igmpproxy/issues/28)**

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -396,7 +396,7 @@ struct SubnetList *parseSubnetAddress(char *addrstr) {
     }
 
     tmpSubnet = (struct SubnetList*) malloc(sizeof(struct SubnetList));
-    tmpSubnet->subnet_addr = addr;
+    tmpSubnet->subnet_addr = (addr & mask);
     tmpSubnet->subnet_mask = ntohl(mask);
     tmpSubnet->next = NULL;
 
--- a/src/config.c
+++ b/src/config.c
@@ -396,7 +396,7 @@ struct SubnetList *parseSubnetAddress(char *addrstr) {
     }
 
     tmpSubnet = (struct SubnetList*) malloc(sizeof(struct SubnetList));
-    tmpSubnet->subnet_addr = (addr & mask);
+    tmpSubnet->subnet_addr = addr;
     tmpSubnet->subnet_mask = ntohl(mask);
     tmpSubnet->next = NULL;
 
--- a/src/ifvc.c
+++ b/src/ifvc.c
@@ -201,8 +201,10 @@ void rebuildIfVc () {
 **
 */
 void buildIfVc(void) {
-    struct ifreq IfVc[ sizeof( IfDescVc ) / sizeof( IfDescVc[ 0 ] )  ];
+    struct ifreq IfVc[ MAX_IF * 2 ]; /* allocate in double in order to get as much information as possible */
     struct ifreq *IfEp;
+    struct IfDesc *dp;
+    struct SubnetList *allowednet, *currsubnet;
     struct Config *config = getCommonConfig();
 
     int Sock;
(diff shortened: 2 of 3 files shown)


ljpaff:
#

src/config.c
```diff
@@ -396,7 +396,7 @@ struct SubnetList *parseSubnetAddress(char *addrstr) {
     }
 
     tmpSubnet = (struct SubnetList*) malloc(sizeof(struct SubnetList));
-    tmpSubnet->subnet_addr = addr;
+    tmpSubnet->subnet_addr = (addr & mask);
```
pali: Why is this change needed?

src/ifvc.c
```diff
@@ -274,6 +281,21 @@ void buildIfVc(void) {
             mask = s_addr_from_sockaddr(&IfReq.ifr_addr); // Do not use ifr_netmask as it is not available on freebsd
             subnet = addr & mask;
 
+            dp = getIfByName(IfPt->ifr_name, 1);
```
pali, line 284: What is the purpose of function parameter which is always constant?

pali (commented):
I have not done review of code yet, just quick look and ...

src/config.c
ljpaff: Because in ifvc.c are now handling subnets on the form addr & mask so it must be changed in order to work properly.

src/ifvc.c
ljpaff: It is to be able to take advantage of the function so that it also works in the same way as it did in the past.

src/config.c
rbgarga: I just would like to note that this specific change was reverted on pfSense

https://github.com/pfsense/FreeBSD-ports/commit/009d60cab48f85b0e75a6a35bd7805ece19e6806

ljpaff: Ok, so this change must be reverted....

ljpaff: You are right, yesterday night i had compiled the source code and tested it. It is working perfectly without this change, and i can confirm that igmpproxy is working with a vlan tagged with pppoe interface.

What are next steps?. Do I cancel this pull request and open another one without config.c change or Do I revert the change in this pull request?

pali: Please update *this* pull request. There is no need to create a new one.

src/ifvc.c
```diff
@@ -201,8 +201,10 @@ void rebuildIfVc () {
 **
 */
 void buildIfVc(void) {
-    struct ifreq IfVc[ sizeof( IfDescVc ) / sizeof( IfDescVc[ 0 ] )  ];
+    struct ifreq IfVc[ MAX_IF * 2 ]; /* allocate in double in order to get as much information as possible */
```
pali, line 204: Why `*2`? Maximal number is `maximal`. So either defined constant is not so large or there is another hidden problem?

src/ifvc.c
```diff
@@ -246,6 +248,19 @@ void buildIfVc(void) {
...
+            if ( IfDescEp >= &IfDescVc[ MAX_IF ] ) {
+                my_log(LOG_DEBUG, 0, "Too many interfaces, skipping all since %s", IfPt->ifr_name);
+                break;
+            }
+
+            /* don't retrieve any info from non-IPv4 interfaces
+             */
+            if ( IfPt->ifr_addr.sa_family != AF_INET ) {
+                my_log(LOG_DEBUG, 0, "Interface is not AF_INET, skipping %s (family %d)", IfPt->ifr_name, IfPt->ifr_addr.sa_family);
+                continue;
```
pali, line 262: Skipping interfaces without IPv4 address would mean igmpproxy itself would not see this interface. Why such change is needed? Previously these interfaces was normally processed just marked as non-IPv4.

src/ifvc.c
```diff
@@ -274,6 +281,21 @@ void buildIfVc(void) {
...
+                allowednet = (struct SubnetList *)malloc(sizeof(struct SubnetList));
+                if (allowednet == NULL) my_log(LOG_ERR, 0, "Out of memory !");
+                allowednet->next = NULL;
+                allowednet->subnet_mask = mask;
+                allowednet->subnet_addr = subnet;
+                currsubnet = dp->allowednets;
+                while (currsubnet->next != NULL)
+                    currsubnet = currsubnet->next;
+                currsubnet->next = allowednet;
+                continue;
```
pali, line 295: This completely skip code which is below. Why?

Looks like this is a hack which just hide the fact that there is a probably bug in code below and try to invisibly skip it. This change is non-intuitive and introduce (potentially) dead code. Also probably may introduce another bugs as parts like `IfDescEp++;` are not executed...

pali (changes requested):

src/ifvc.c
pali: But that function is not called from other places. Such change introduce a dead code as parameter is never false.

pali:
@ljpaff Could you please address issues above?

PR #40, opened 2018-08-07
Feature/code clean

ghost:
- make more stuff static
- keep some defines local
- move some globals where they belong


src/confread.c
```diff
@@ -51,40 +51,42 @@
...
-char    cToken[MAX_TOKEN_LENGTH];   // Token buffer...
-short   validToken;
+static struct {
+    FILE            *confFilePtr;       // File handle pointer
+    char            *iBuffer;           // Inputbuffer for reading...
+    unsigned int    bufPtr;             // Buffer position pointer.
+    unsigned int    readSize;           // Number of bytes in buffer after last read...
+    char    cToken[MAX_TOKEN_LENGTH];   // Token buffer...
+    short   validToken;
+} conf;
```
pali: Is putting all variables into struct really needed? They are private for confread.c file, so `static` should be enough.

pali (commented):
You can squash two commits which adds `static` keyword into one commit.

src/confread.c
ghost: ok

pali (changes requested):
Seems that this pull request is broken as it contains broken merge. Please rebase and fix it.

ghost:

PR #41, opened 2018-08-07
clean/simplify mcgroup

ghost:
- prefix module functions
- keep udp socket local
- simplify static common function


pali:
I'm not a big fan of mixing upper-case and camel-case in one identifier. Also you should write commit messages more descriptive (see history for examples).

ghost:
do you like "mcgroup_sendjoin"?

pali:
Look, majority of igmpproxy code uses "camelCase" style with leading lowercase character.

ghost:
Good, in this case a module prefix wont make sense i guess.
What about moving the function getMcGroupSock from rttable.c to mcgroup.c (intention to make it local/static)? I could make a separate pull request for that.

pali:
So show code and then we can discuss about it.

ghost:

PR #42, opened 2018-08-07
Feature/refactor callout

ghost:
- rename callout to timer
- drop unused functions

This is the first part of what i plan to do here. What i would like to do as well is the whole
"handle timeouts"section  in igmpproxy.c should be moved to timer.c. And the implemenation of aging is very complex and can be heavily simplified. I will do this afterwards.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -1,6 +1,6 @@
 sbin_PROGRAMS = igmpproxy
 igmpproxy_SOURCES = \
-	callout.c \
+	timer.c \
 	config.c \
 	confread.c \
 	ifvc.c \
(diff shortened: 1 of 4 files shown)


pali:
Commit message says `rename callout.c to timer.c and adapt exposed function accordingly` but there is no explanation.

ghost:
i can come up with a better commit message: "rename callout.c to timer.c to better match the functionality inside the module and there are multiple functions inside, that start with timer_ already"

pali:
Look! I would suggest you to read how to write commit messages. If first line of commit message spans 3 lines, something is really wrong.

Some useful sources about commit messages:
https://www.kernel.org/doc/html/latest/process/submitting-patches.html#describe-your-changes
https://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines#New_Development
https://github.com/torvalds/subsurface/blob/a48494d2fbed58c751e9b7e8fbff88582f9b2d02/README#L88

I hope this would help you.

ghost:

PR #43, opened 2018-08-07
Feature/clean igmp

ghost:
- prefix functions
- move recvfrom into igmp module, this will allow later to keep recv_buf locally

### DIFF
--- a/src/igmp.c
+++ b/src/igmp.c
@@ -32,7 +32,7 @@
 **
 */
 /**
-*   igmp.h - Recieves IGMP requests, and handle them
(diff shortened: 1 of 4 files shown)


pali:
This just introduce mess with camel-case and identifiers with underline. About formatting there is already open, but unfinished pull request: https://github.com/pali/igmpproxy/pull/19

ghost:
ok, i will drop the formatting commits, what about the "move recvfrom part to igmp.c" commit?

ghost:

PR #44, opened 2018-08-08
prefix request functions

ghost:
- prefix exported functions with request_


pali:
Same there. Either camel case or underlines, or upper/lower case. But not mix of everything.

ghost:
ok, will come up with new approach

PR #48, opened 2018-11-23
Forward MembershipQuery from upstream to downstream

aimwang:
If the conf->queryInterval was much larger than upstream. The igmp snooping will kick out the membership of igmpproxy member. This commit forward upstream MembershipQuery to downstream, ensure that the MembershipReport can be delivered in time.

### DIFF
--- a/src/igmp.c
+++ b/src/igmp.c
@@ -251,7 +251,7 @@ void acceptIgmp(int recvlen) {
         return;
 
     case IGMP_MEMBERSHIP_QUERY:
-		acceptMembershipQuery(src);
+        acceptMembershipQuery(src);
         return;
 
     default:
--- a/src/igmp.c
+++ b/src/igmp.c
@@ -251,6 +251,7 @@ void acceptIgmp(int recvlen) {
         return;
 
     case IGMP_MEMBERSHIP_QUERY:
+		acceptMembershipQuery(src);
         return;
 
     default:
--- a/src/igmpproxy.c
+++ b/src/igmpproxy.c
@@ -273,7 +273,7 @@ void igmpProxyRun(void) {
     lasttime = curtime;
 
     // First thing we send a membership query in downstream VIF's...
-    sendGeneralMembershipQuery();
+    sendGeneralMembershipQuery(1);
 
     // Loop until the end...
     for (;;) {
--- a/src/igmpproxy.c
+++ b/src/igmpproxy.c
@@ -273,7 +273,7 @@ void igmpProxyRun(void) {
     lasttime = curtime;
 
     // First thing we send a membership query in downstream VIF's...
-    sendGeneralMembershipQuery(1);
+    sendGeneralMembershipQuery();
 
     // Loop until the end...
     for (;;) {
--- a/src/igmpproxy.h
+++ b/src/igmpproxy.h
@@ -272,7 +272,8 @@ int getMcGroupSock(void);
  */
 void acceptGroupReport(uint32_t src, uint32_t group);
 void acceptLeaveMessage(uint32_t src, uint32_t group);
-void sendGeneralMembershipQuery(void);
+void acceptMembershipQuery(uint32_t src);
+void sendGeneralMembershipQuery(int routine);
 
 /* callout.c 
 */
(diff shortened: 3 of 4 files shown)


pali (changes requested):
Hello @aimwang! Sorry for a late reply, I forgot about it.

Your change looks good, there are just few things for fixing:

1) Indentation & style. You are mixing spaces and tabs. igmpproxy source code uses only spaces, so please fix yout pull request to use consistent style, spaces and correclty indent code.

2) sendGeneralMembershipQuery function. Please split this function into two functions, one loop throw downstream interfaces and one which installs a new timer for sendGeneralMembershipQuery function. So a new parameter for sendGeneralMembershipQuery function is not needed.

aimwang:
I add a new function doSendGeneralMembershipQuery() to send query to downstream interfaces. When receive MembershipQuery will only send query immediately, and do not change timer of sendGeneralMembershipQuery().

pali:
Ok, thank you! Now it is more readable.

But I'm thinking about this logic and now I'm not sure if this change is really correct.

igmp query and report packets between upstream router and router on which is running igmpproxy are processed by kernel. igmpproxy does not process them at all, nor it does not generate any report membership packets.

So this change looks like it does not do anything, just more frequenly send query memebership packets to downstream network. Which just not fix anything related to upstream network.

So what is the problem which you are trying to fix? I do not fully understand what you mean by `The igmp snooping will kick out the membership of igmpproxy member.`

aimwang:
Media Server MembershipQuery Interval = 10 (Not default but exist)
IGMP Proxy   MembershipQuery Interval = 60
The media server send MembershipQuery but ignored by IGMP Proxy.
Media will stop streaming when wait timeout.
~~~
--- : signal
=== : stream
Sec    Media Server           IGMP Proxy              VLC
---    ------------           ----------              ---
0      |                      | <-MembershipReport    |
1      | <-MembershipReport   |                       |
2      | ===================> | ====================> |
10     | Membership Query->   |                       |
11-59  | (Stop streaming)     |                       |
60     |                      | Membership Query->    |
61     |                      | <-MembershipReport    |
62     | <-MembershipReport   |                       |
63     | ===================> | ====================> |
       |                      |                       |
~~~
There are three way to avoid this situation. This patch was pick sol 1.
1. Forward membershi query to down stream. When VLC report membership, igmpproxy will forward to upstream.
2. Keep all membership status and generate membership report by igmp proxy.
3. Force Media Server set membershi query interval greater then 60 sec.

pali:
> 1. Forward membershi query to down stream. When VLC report membership, igmpproxy will forward to upstream.

igmpproxy does *not* send IGMP MembershipReport packets to Media Server. You can verify source code that `sendIgmp` is called only with `IGMP_MEMBERSHIP_QUERY` type.

It is kernel who send and receive packets between Media Server and IGMP Proxy in your diagram.

> 2. Keep all membership status and generate membership report by igmp proxy.

This is already implemented, but in kernel. And kernel's solution should be used.

igmpproxy just calls setsockopt with IP_ADD_MEMBERSHIP option, but this does only reference counter in kernel. This option is added by joinMcGroup() call and this function is not even called in your patch. So I really doubt that this patch implements solution 1. as you described.

Linux kernel automatically send MembershipReport packets when it receive Membership Query packets from upstream (Media Server). So if it does not work for you I think you hit some kernel bug. Are you using Linux kernel? Or any other?

Also on Linux systems it can be common problem that iptables is incorrectly configured which cause igmpproxy packets are dropped before they arrive to kernel's code which handle Membership Query. igmpproxy uses RAW sockets which bypass kernel's iptables rules and therefore igmpproxy receive also packets which are explicitly dropped by iptables. So it is possible that your patch receive Membership Query packets even when you configured iptables to drop them. Can you check if this is not your problem?

`iptables -I INPUT -p igmp -j ACCEPT` and `iptables -I OUTPUT -p igmp -j ACCEPT`

adds iptables rule to accept all IGMP packets.

aimwang:
Hi Pali,

Thanks for your tips. I will find out why my kernel did not rejoin_groups when it received Membership_query. thank you very much.

PR #59, opened 2020-03-02
Fix an error for timer execute 

onesecret:
If not fix , this timer execute every 10s or more。
The variable secs must be assigned and less than 3.Because the timer cycle each 3 seconds.

### DIFF
--- a/src/igmpproxy.c
+++ b/src/igmpproxy.c
@@ -338,7 +338,7 @@ void igmpProxyRun(void) {
              * call gettimeofday.
              */
             if (Rt == 0) {
-                curtime.tv_sec = lasttime.tv_sec + secs;
+                curtime.tv_sec = lasttime.tv_sec + secs;// secs must less than 3
                 curtime.tv_nsec = lasttime.tv_nsec;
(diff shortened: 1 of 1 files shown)


pali:
Could you please describe problem/error more deeply? After quick look I do not see what is the problem.

onesecret:
at line 295:
        secs = timer_nextTimer();//here secs be assined, it can be 125 , 10, or others.its value depends on timer's queue->time.
if not alter secs value, at line 341:
curtime.tv_sec = lasttime.tv_sec + secs;
curtime.tv_sec will too large, and cause difftime.tv_sec too large,  
difftime.tv_sec is a param of function age_callout_queue.
at this functioon, this param is compared with queue->time. if this param is larger than queue->time, this timer will be executed immediately.
Actually, this param means how long the real time has passed.
Did i make it clear ?

pali:
Yes! I understood what you mean and how your change fixes (or rather workarounds?) this problem.

I wrote about this issue also in comment: https://github.com/pali/igmpproxy/pull/32#issuecomment-620887757

Issue #62, opened 2020-04-26
MAX_IF 40 not sufficient for freebsd


Uglymotha:
This has been reported before I think, but because freebsd creates multiple logical interfaces for one actual, the default value of 40 for MAX_IF is often not sufficient. In my case even 200 was not enough, since my system has 16 phys ints, several laggs, 10 vlans per lagg and bridges across these. 

I created the attached patch to fix this. I can create a pull request if you are willing to accept the patch.

Uglymotha:
[freebsd_max_if.zip](https://github.com/pali/igmpproxy/files/4535067/freebsd_max_if.zip)

pali:
I know about this issue, but your patch does not fix it. It just increase max limit to some random number. Real fix for this problem is to remove upper limit and change static memory allocation to dynamic one based on number of interfaces in system.

Uglymotha:
Shouldn't that be fairly simple to acheive by using getifaddrs() / freeifaddrs() ?
These functions already use dynamic memory allocation.

pali:
Well, I have not looked at it deeply how to implement it. If you this it is simple, feel free to open a pull request.

Uglymotha:
First off, it's been some 20-25 years since I've done any real programming in c. Been looking through and analysing the source code yesterday and igmpproxy seems a basic enough project to get back into it. Like many people I currently have some time free to do this. So here's what I now think about solution to this issue.

1. Move the global definition of the IfDescVC[] array and *fDescEp into main(). It's good programming practice anyway.
2. Initialize the array at start-up based on the actual nr of interfaces in the system. The amount of memory needed will be known, also static (have to analyse the RebuildIfVc function yet to see how dynamic the array might be there)
3. Pass the pointers as arguments to the functions that need them. 

Functions in ifvc.c are not used that much and I think I already have much of the above covered.

Uglymotha:
Oh and maybe have MAX_IF be configurable. igmpproxy is often run on memory restrained systems

pali:
> Oh and maybe have MAX_IF be configurable. igmpproxy is often run on memory restrained systems

`struct IfDesc` is 50-60 bytes long. You would have to about 100 000 interfaces to hit memory limit problems (10 MB) on small memory-limited systems. But I guess your memory-limited system would not be able to process so much interfaces as cost for it would be larger then cost for for igmpproxy.

[...]

Issue #58, opened 2020-02-29
igmpproxy timer execute error


onesecret:
diff --git a/src/igmpproxy.c b/src/igmpproxy.c
index 9d1172b..46a659b 100644
--- a/src/igmpproxy.c
+++ b/src/igmpproxy.c
@@ -297,7 +297,7 @@ void igmpProxyRun(void) {
             timeout = NULL;
         } else {
             timeout->tv_nsec = 0;
**-            timeout->tv_sec = (secs > 3) ? 3 : secs; // aimwang: set max timeout**
**+            timeout->tv_sec = secs = (secs > 3) ? 3 : secs; // aimwang: set max timeout**
         }
 
         // Prepare for select.

it cause a problem at :

        // At this point, we can handle timeouts...
        do {
            /*
             * If the select timed out, then there's no other
             * activity to account for and we don't need to
             * call gettimeofday.
             */
            if (Rt == 0) {
                curtime.tv_sec = lasttime.tv_sec + secs;
                curtime.tv_nsec = lasttime.tv_nsec;
                Rt = -1; /* don't do this next time through the loop */
            } else {
                clock_gettime(CLOCK_MONOTONIC, &curtime);
            }
            difftime.tv_sec = curtime.tv_sec - lasttime.tv_sec;
            difftime.tv_nsec += curtime.tv_nsec - lasttime.tv_nsec;
            while (difftime.tv_nsec > 1000000000) {
                difftime.tv_sec++;
                difftime.tv_nsec -= 1000000000;
            }
            if (difftime.tv_nsec < 0) {
                difftime.tv_sec--;
                difftime.tv_nsec += 1000000000;
            }
            lasttime = curtime;
            if (secs == 0 || difftime.tv_sec > 0)
                age_callout_queue(difftime.tv_sec);
            secs = -1;
        } while (difftime.tv_sec > 0);

onesecret:
I want to push a commit for it , but I don't how to do .
it likes this:

onesecret@onesecret-Vostro-3470:~/igmpproxy$ git push origin HEAD:refs/for/master
Username for 'https://github.com': onesecret
Password for 'https://onesecret@github.com': 
remote: Permission to pali/igmpproxy.git denied to onesecret.
fatal: unable to access 'https://github.com/pali/igmpproxy.git/': The requested URL returned error: 403

can I push a commit ? It looks like I have no permission.

pali:
[...]

PR #63, opened 2020-04-30
Dynamic IfDesc Table

Uglymotha:
Fixes issue #62 and build errors on freebsd because of incorrect placement of includes.

### DIFF
--- a/doc/igmpproxy.conf.5.in
+++ b/doc/igmpproxy.conf.5.in
@@ -3,7 +3,7 @@
 .\" .I  - green or kursive (on HTML)
 .\" .TP - paragraph ? (header line, followed by indented lines)
 .\"
-.TH igmpproxy.conf 5 "" "@PACKAGE_STRING@"
+.TH igmpproxy.conf 5 "" "igmpproxy 0.2.1"
 .SH NAME
 igmpproxy.conf \- Configuration file for
 .BR igmpproxy (8)
@@ -11,7 +11,7 @@ multicast daemon
 
 .SH DESCRIPTION
 .B igmpproxy.conf
-contains the configuration for the 
+contains the configuration for the
 .B igmpproxy
 multicast daemon. It defines which network interfaces should be
 used by the routing daemon. Each interface must be give one of the following roles:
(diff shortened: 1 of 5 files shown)


pali:
Ok, so you have already figured out that there is automatic compile check for every pull request and you have already fixed code as to be compatible with more compiler versions...

Some more comments:
* If diff there are lot of whitespace changes, which make only noise and make reviewing patches more complicated. If there are no reasons for whitespace changes, please remove them.
* Squash "fixup" commits into some relevant previous commit (see `git rebase --interactive`) and then do force update (see `git push --force`); every commit in history should be compile-able
* If it is not really necessary, do not try to change indentation of code as it just increase diff

Uglymotha:
Over the last few days my rework of the code has become increasingly extensive. Igmpproxy was indeed a good project to get back into programming, it's definitely been fun. 
In some places the code was just ancient, like using socket io to get interface info instead of getifaddrs(). In other places  code was shabby, in some places bad and in a couple absolutely terrible, like the main loop and queue timing. The project seems to hav esuffered from repeated unfinished attempts to implement functionality. Leading to a bit of a mess. Outside of the actual igmp functionality it's all redone the way it should be. Lots a functions were altered to more readable and flexible code, some functionality split of into new functions and some removed.

I currently have the interface and configuration management working correctly as far as I am able to test, SIGHUP is functional and main loop timing is now implemented in just 12 simple lines of code, using 1 call to clock_gettime. instead of the some 60 lines it was before. This also fixes issue #58 without adding another shabby patch to a completely incorrectly implemented timing.

I am now reworking the callout queue, when that's finished I'll do some more testing of my own before I am confident in releasing the new code. Hopefully I'll be able to get a version of this all in my own repo today.

Leaves the question of how to proceed. Opening a new pull request for all the changes when I'm finished or maybe fork off completely. My preference would be the first.

pali:
Please separate changes (ideally to more pull reaquests) as reviewing one big change is hard. Also please update this pull request as I mentioned in previous comment so I can look at it.

Uglymotha:
After some more testing, I am now confident to release the new code base.

Closing this PR and opening a new one, is it has become much more than just dynamic ifdesc tables.

Issue #35, opened 2018-05-30
Can't install from source (FreeBSD 11.1)


oasisck:
Hi, on fresh install of FreeBSD 11.1 DVD
```
root@freebsd:~/igmpproxy-0.2.1 # uname -a
FreeBSD freebsd 11.1-RELEASE FreeBSD 11.1-RELEASE #0 r321309: Fri Jul 21 02:08:28 UTC 2017     root@releng2.nyi.freebsd.org:/usr/obj/usr/src/sys/GENERIC  amd64
```
Got those errors
```
root@freebsd:~/igmpproxy-0.2.1 # make
make  all-recursive
Making all in doc
Making all in src
cc -DHAVE_CONFIG_H -I. -I..      -g -O2 -MT callout.o -MD -MP -MF .deps/callout.Tpo -c -o callout.o callout.c
In file included from callout.c:36:
In file included from ./igmpproxy.h:39:
In file included from ./os.h:1:
In file included from /usr/include/net/route.h:36:
/usr/include/sys/counter.h:32:9: error: unknown type name 'uint64_t'
typedef uint64_t *counter_u64_t;
        ^
In file included from callout.c:36:
In file included from ./igmpproxy.h:39:
In file included from ./os.h:1:
/usr/include/net/route.h:59:2: error: unknown type name 'uint16_t'
        uint16_t        ro_plen;
        ^
/usr/include/net/route.h:60:2: error: unknown type name 'uint16_t'
        uint16_t        ro_flags;
        ^
/usr/include/net/route.h:61:2: error: unknown type name 'uint16_t'
        uint16_t        ro_mtu; /* saved ro_rt mtu */
        ^
/usr/include/net/route.h:62:2: error: unknown type name 'uint16_t'
        uint16_t        spare;
        ^
/usr/include/net/route.h:63:18: error: field has incomplete type 'struct sockaddr'
        struct  sockaddr ro_dst;
                         ^
/usr/include/net/route.h:63:9: note: forward declaration of 'struct sockaddr'
        struct  sockaddr ro_dst;
                ^
/usr/include/net/route.h:82:2: error: unknown type name 'u_long'; did you mean 'long'?
        u_long  rmx_locks;      /* Kernel must leave these values alone */
        ^
/usr/include/net/route.h:83:2: error: unknown type name 'u_long'; did you mean 'long'?
        u_long  rmx_mtu;        /* MTU for this path */
        ^
[...]

PR #66, opened 2020-05-08
Changes to timing

Uglymotha:
So here's the first new pull request with changes to the timing structure. This fixes issue #59 and build errors on freebsd due to incorrect placement of includes in igmpproxy.c and igmpproxy.h

To make this compilable and functional I also have to introduce a couple of (as of yet) unused functions. These are reloadConfig() and createVifs(), a function that will be used to dynamically update and create vifs during config reload. JoinMcRouterGroup(), a split of funcionality from initRouteTable(), also to be used for dynamically managed interfaces. The function clearAllRoutes was changed to clearRoutes, it will also be used in dynamice interfaces to remove routes for a certain upstream if, hence the name change.

Some new structs are also defined and some change because they are used as such in the new functions.

### DIFF
--- a/src/callout.c
+++ b/src/callout.c
@@ -40,11 +40,11 @@ static int id = 0;
 static struct timeOutQueue  *queue = NULL; /* pointer to the beginning of timeout queue */
 
 struct timeOutQueue {
-    struct timeOutQueue    *next;   // Next event in queue
     int                     id;
     timer_f                 func;   // function to call
     void                    *data;  // Data for function
     long                    time;   // Time for event
+    struct timeOutQueue     *next;  // Next event in queue
 };
 
 // Method for dumping the Queue to the log.
@@ -56,10 +56,8 @@ static void debugQueue(void);
 void free_all_callouts(void) {
     struct timeOutQueue *p;
 
-    while (queue) {
-        p = queue;
-        queue = queue->next;
-        free(p);
+    for (p = queue ? queue->next : NULL; queue; queue = p, p = queue->next) {
+        free(queue);   // Alloced by timer_setTimer()
     }
 }
 
@@ -78,7 +76,7 @@ void age_callout_queue(struct timespec curtime) {
             ptr->func(ptr->data);
         }
         queue = ptr = ptr->next;
-        free(tmp);
+        free(tmp);  // Alloced by timer_setTimer()
         i++;
     }
 }
@@ -94,9 +92,9 @@ int timer_setTimer(int delay, timer_f action, void *data) {
     struct timespec curtime;
     int i = 1;
 
-    // create a node.
+    // create a node. Freed by free_all_callouts() and age_callout_queue().
     node = (struct timeOutQueue *)malloc(sizeof(struct timeOutQueue));
-    if (node == 0) {
+    if (! node) {
         my_log(LOG_WARNING, 0, "Malloc Failed in timer_settimer\n");
         return -1;
     }
@@ -104,24 +102,20 @@ int timer_setTimer(int delay, timer_f action, void *data) {
     node->func = action;
     node->data = data;
     node->time = curtime.tv_sec + delay;
-    node->next = NULL;
     node->id   = ++id;
 
-    // if the queue is empty, insert the node and return.
-    if (!queue) queue = node;
-    else {
+    if (! queue) {
+        // if the queue is empty, insert the node and return.
+        queue = node;
+    } else {
         // chase the queue looking for the right place. 
         for (i++; ptr->next && node->time >= ptr->next->time; ptr = ptr->next, i++);
         if (ptr == queue && node->time < ptr->time) {
            // Start of queue, insert.
            queue = node;
            node->next = ptr;
-        } else if ( ptr->next ) {
-           // Middle of queue, insert
-           node->next = ptr->next;
-           ptr->next = node;
         } else {
-           // End of queue, append.
+           node->next = ptr->next;
            ptr->next = node;
         }
     }
@@ -134,16 +128,15 @@ int timer_setTimer(int delay, timer_f action, void *data) {
 *   returns the time until the timer is scheduled
 */
 int timer_leftTimer(int timer_id) {
-    struct timeOutQueue *ptr = queue;
+    struct timeOutQueue *ptr;
     struct timespec curtime;
 
     if (!timer_id || !queue) return -1;
-    while (ptr && ptr->id != timer_id) {
-        ptr = ptr->next;
-    }
-    if (ptr){
-        clock_gettime (CLOCK_MONOTONIC, &curtime);
-        return (ptr->time - curtime.tv_sec);
+    for (ptr = queue; ptr; ptr = ptr->next) {
+        if (ptr->id == timer_id) {
+            clock_gettime(CLOCK_MONOTONIC, &curtime);
+            return (ptr->time - curtime.tv_sec);
+        }
     }
     return -1;
 }
--- a/src/callout.c
+++ b/src/callout.c
@@ -71,10 +71,12 @@ void age_callout_queue(struct timespec curtime) {
     int i = 1;
  
     if (curtime.tv_sec == 0) clock_gettime (CLOCK_MONOTONIC, &curtime);
-    while (ptr && ((ptr->time <= curtime.tv_sec) || ( curtime.tv_nsec >= 500000000 && ptr->time <= curtime.tv_sec-1))) {
+    while (ptr && ((ptr->time <= curtime.tv_sec) || (curtime.tv_nsec >= 500000000 && ptr->time <= curtime.tv_sec-1))) {
         my_log(LOG_DEBUG, 0, "About to call timeout %d (#%d)", ptr->id, i);
         struct timeOutQueue *tmp = ptr;
-        if (ptr->func) ptr->func(ptr->data);
+        if (ptr->func) {
+            ptr->func(ptr->data);
+        }
         queue = ptr = ptr->next;
         free(tmp);
         i++;
@@ -109,13 +111,15 @@ int timer_setTimer(int delay, timer_f action, void *data) {
     if (!queue) queue = node;
     else {
         // chase the queue looking for the right place. 
-        for ( i++; ptr->next && node->time >= ptr->next->time; ptr = ptr->next ) i++;
-        if ( ptr == queue && node->time < ptr->time ) {
+        for (i++; ptr->next && node->time >= ptr->next->time; ptr = ptr->next, i++);
+        if (ptr == queue && node->time < ptr->time) {
            // Start of queue, insert.
-           queue = node; node->next = ptr;
+           queue = node;
+           node->next = ptr;
         } else if ( ptr->next ) {
            // Middle of queue, insert
-           node->next = ptr->next; ptr->next = node;
+           node->next = ptr->next;
+           ptr->next = node;
         } else {
            // End of queue, append.
            ptr->next = node;
@@ -134,8 +138,10 @@ int timer_leftTimer(int timer_id) {
     struct timespec curtime;
 
     if (!timer_id || !queue) return -1;
-    while (ptr && ptr->id != timer_id) ptr = ptr->next;
-    if ( ptr ){
+    while (ptr && ptr->id != timer_id) {
+        ptr = ptr->next;
+    }
+    if (ptr){
         clock_gettime (CLOCK_MONOTONIC, &curtime);
         return (ptr->time - curtime.tv_sec);
     }
@@ -146,10 +152,10 @@ int timer_leftTimer(int timer_id) {
  * debugging utility
  */
 static void debugQueue(void) {
-    struct timeOutQueue  *ptr; int i = 1;
+    struct timeOutQueue  *ptr; 
+    int i = 1;
 
-    for (ptr = queue; ptr; ptr = ptr->next) {
+    for (ptr = queue; ptr; ptr = ptr->next, i++) {
         my_log(LOG_DEBUG, 0, "(%d - Id:%d, Time:%d) ", i, ptr->id, ptr->time);
-        i++;
     }
 }
--- a/src/callout.c
+++ b/src/callout.c
@@ -153,9 +153,9 @@ int timer_leftTimer(int timer_id) {
  */
 static void debugQueue(void) {
     struct timeOutQueue  *ptr; 
-    int i = 1;
+    int i;
 
-    for (ptr = queue; ptr; ptr = ptr->next, i++) {
+    for (i = 1, ptr = queue; ptr; ptr = ptr->next, i++) {
         my_log(LOG_DEBUG, 0, "(%d - Id:%d, Time:%d) ", i, ptr->id, ptr->time);
     }
 }
--- a/src/callout.c
+++ b/src/callout.c
@@ -146,7 +146,7 @@ int timer_leftTimer(int timer_id) {
  * debugging utility
  */
 static void debugQueue(void) {
-    struct timeOutQueue  *ptr; 
+    struct timeOutQueue  *ptr;
     int i;
 
     for (i = 1, ptr = queue; ptr; ptr = ptr->next, i++) {
--- a/src/callout.c
+++ b/src/callout.c
@@ -103,6 +103,7 @@ int timer_setTimer(int delay, timer_f action, void *data) {
     node->data = data;
     node->time = curtime.tv_sec + delay;
     node->id   = ++id;
+    node->next = NULL;
 
     if (! queue) {
         // if the queue is empty, insert the node and return.
--- a/src/callout.c
+++ b/src/callout.c
@@ -37,25 +37,18 @@
 
 /* the code below implements a callout queue */
 static int id = 0;
-static struct timeOutQueue  *queue = 0; /* pointer to the beginning of timeout queue */
+static struct timeOutQueue  *queue = NULL; /* pointer to the beginning of timeout queue */
 
 struct timeOutQueue {
     struct timeOutQueue    *next;   // Next event in queue
     int                     id;
     timer_f                 func;   // function to call
     void                    *data;  // Data for function
-    int                     time;   // Time offset for next event
+    long                    time;   // Time for event
 };
 
 // Method for dumping the Queue to the log.
 static void debugQueue(void);
-
-/**
-*   Initializes the callout queue
-*/
-void callout_init(void) {
-    queue = NULL;
-}
 
 /**
 *   Clears all scheduled timeouts...
@@ -70,58 +63,22 @@ void free_all_callouts(void) {
     }
 }
 
-
 /**
- * elapsed_time seconds have passed; perform all the events that should
- * happen.
+ *  Execute all expired timers, using .5s grace.
  */
-void age_callout_queue(int elapsed_time) {
-    struct timeOutQueue *ptr;
-    struct timeOutQueue *_queue = NULL;
-    struct timeOutQueue *last = NULL;
-    int i = 0;
-
-    for (ptr = queue; ptr; ptr = ptr->next) {
-        if (ptr->time > elapsed_time) {
-            ptr->time -= elapsed_time;
-            break;
-        } else {
-            elapsed_time -= ptr->time;
-            if (_queue == NULL)
-                _queue = ptr;
-            last = ptr;
-         }
-    }
-
-    queue = ptr;
-    if (last) {
-        last->next = NULL;
-    }
-
-    /* process existing events */
-    for (ptr = _queue; ptr; ptr = _queue, i++) {
-        _queue = _queue->next;
+void age_callout_queue(struct timespec curtime) {
+    struct timeOutQueue *ptr = queue;
+    int i = 1;
+ 
+    if (curtime.tv_sec == 0) clock_gettime (CLOCK_MONOTONIC, &curtime);
+    while (ptr && ((ptr->time <= curtime.tv_sec) || ( curtime.tv_nsec >= 500000000 && ptr->time <= curtime.tv_sec-1))) {
         my_log(LOG_DEBUG, 0, "About to call timeout %d (#%d)", ptr->id, i);
-        if (ptr->func)
-             ptr->func(ptr->data);
-        free(ptr);
+        struct timeOutQueue *tmp = ptr;
+        if (ptr->func) ptr->func(ptr->data);
+        queue = ptr = ptr->next;
+        free(tmp);
+        i++;
     }
-}
-
-/**
- * Return in how many seconds age_callout_queue() would like to be called.
- * Return -1 if there are no events pending.
- */
-int timer_nextTimer(void) {
-    if (queue) {
-        if (queue->time < 0) {
-            my_log(LOG_WARNING, 0, "timer_nextTimer top of queue says %d", 
-                queue->time);
-            return 0;
-        }
-        return queue->time;
-    }
-    return -1;
 }
 
 /**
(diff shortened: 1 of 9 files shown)


src/callout.c
```diff
@@ -70,58 +63,22 @@ void free_all_callouts(void) {
...
-
-    /* process existing events */
-    for (ptr = _queue; ptr; ptr = _queue, i++) {
-        _queue = _queue->next;
+void age_callout_queue(struct timespec curtime) {
+    struct timeOutQueue *ptr = queue;
+    int i = 1;
+ 
+    if (curtime.tv_sec == 0) clock_gettime (CLOCK_MONOTONIC, &curtime);
+    while (ptr && ((ptr->time <= curtime.tv_sec) || ( curtime.tv_nsec >= 500000000 && ptr->time <= curtime.tv_sec-1))) {
```
pali, line 74: What does 500000000s means?

pali: Now It see that it is 0.5s

src/config.c
```diff
@@ -260,6 +292,97 @@ void configureVifs(void) {
...
+                if (Dp->Flags & IFF_LOOPBACK) continue;
+            } else {
+                // New Interface. Only add valid up/downstream vif.
+                if ( (Dp->Flags & IFF_LOOPBACK) || (Dp->state != IF_STATE_DOWNSTREAM && Dp->state != IF_STATE_UPSTREAM) ) continue;
+                oDp=NULL;
+            }
+        }
+        if(Dp->state == IF_STATE_UPSTREAM) {
+            if (upsvifcount < MAX_UPS_VIFS -1)
+            {
```
pali: Just style comment: put `{` on previous line and change swap if-else braches so you can compare `upsvifcount` and `MAX_UPS_VIFS` without `-1`:
```c
if (upsvifcount >= MAX_UPS_VIFS) {
/* report error */
} else {
/* OK */
}
```
`+1` and `-1` are lot of times suspicous.

src/igmpproxy.c
```diff
@@ -38,8 +38,10 @@
 */
 
 /* getopt() and clock_getime() */
-#ifndef _POSIX_C_SOURCE
-#define _POSIX_C_SOURCE 200112L
+#ifndef __FreeBSD__
+    #ifndef _POSIX_C_SOURCE
+    #define _POSIX_C_SOURCE 200112L
+    #endif
 #endif
```
pali: This change is suspicous. _POSIX_C_SOURCE 200112 is needed for getopt() and clock_getime() functions. I guess this #ifndef is some hack to workaround some problem. It should be fixed properly. Moreover it is not related to timing code rewrite, so should be in seprate commit.

src/igmpproxy.h
```diff
@@ -35,8 +35,10 @@
 *   igmpproxy.h - Header file for common includes.
 */
 
-#include "config.h"
-#include "os.h"
+#ifndef __FreeBSD__
+    #include "config.h"
```
pali: `config.h` must be always included and must be included as the first file. This is required for autoconfig/automake system. Disabling it inclusion on FreeBSD just makes autoconfig/automake broken.

src/igmpproxy.h
```diff
@@ -60,6 +62,11 @@
 #include <netinet/in.h>
 #include <arpa/inet.h>
 
+#ifdef __FreeBSD__
+    #include "config.h"
```
pali: Same here, config.h must be included as the first file.

pali:
When writing a new code, try to use consistent style. E.g. put `if (...)` and its following `function()` call into two lines (not as oneliner). Same for `for (...) if (...) break;` put into 3 lines, not one long oneliner. Also in new code try to use consistent spacing for constructions `if (var)`, `for(var...` etc.

src/igmpproxy.h
Uglymotha: https://github.com/pali/igmpproxy/issues/35

You're right about config.h of course, will fix that. 

I thought I would finally include patch for this too.

src/igmpproxy.c
Uglymotha: issue #35

src/igmpproxy.h
Uglymotha: issue #35

src/config.c
Uglymotha: Fixed

Uglymotha:
Can’t find this comment, did you already delete it? It’s of course 500.000.000ns ie 0.5s.

Uglymotha:
> When writing a new code, try to use consistent style. E.g. put if (...) and its following function() call into two lines (not as oneliner). Same for for (...) if (...) break; put into 3 lines, not one long oneliner. Also in new code try to use consistent spacing for constructions if (var), for(var... etc.

I have adapted to the style used in the file, so it may be little bit different, because styles were different across files. 
For oneliners, I really like to use them when the actions involved are short and tied together. Like inserting into a list. I learned programming on the ZX Spectrum and C64. On these memory restrained systems you learn that having code on separate lines adds 2 bytes per line to your code size. :)

Also a logics table with few short actions, like this from createvifs:
                switch (oDp->state) {
                case IF_STATE_DISABLED:
                    switch (Dp->state) {
                    case IF_STATE_DISABLED:   {                                                      continue; }
                    case IF_STATE_DOWNSTREAM: {                                  oDp=NULL;  join=1;  break; }
                    case IF_STATE_UPSTREAM:   {                                  oDp=NULL;           break; }
                    }
                    break;

will be much harder to read if the statements are put on their own line.

src/igmpproxy.c
pali: This change is some nonsense hack. Any explanation why it is needed? And of course should not be part of change which touches timing code.

src/igmpproxy.h
```diff
@@ -36,7 +36,9 @@
 */
 
 #include "config.h"
-#include "os.h"
+#ifndef __FreeBSD__
+    #include "os.h"
+#endif
```
pali: And this is suspicious too. `os.h` is file prepared by autoconf and contains system dependent definitions. It really should be included as soon as possible, therefore after `config.h`. All platform dependent stuff should be done in this `os.h` file and not in platform-independent `igmpproxy.h` file. This is reason why `os.h` exists.

pali:
> Can’t find this comment, did you already delete it? It’s of course 500.000.000ns ie 0.5s.

I already explained on github that it is mistake, it is 0.5s and then marked comment as obsolate. Apparently this explanation was not sent by email.

pali:
> I have adapted to the style used in the file, so it may be little bit different, because styles were different across files.

So please choose one style which is already used in igmpproxy and do not invent a new one. Because the result would be like in XKCD 927.

pali:
About style, look at these your lines:

`if ( ptr == queue && node->time < ptr->time ) {`

`while (ptr && ptr->id != timer_id)`

`if ( ptr ){`

`if(Dp->state == IF_STATE_UPSTREAM) {`

`if (upsvifcount >= MAX_UPS_VIFS) {`

Every one has different style where are put spaces. Please when writing a new code, make this consistent.

src/rttable.c
```diff
@@ -430,7 +442,7 @@ int insertRoute(uint32_t group, int ifx, uint32_t src) {
 *   activated, it's reinstalled in the kernel. If
 *   the route is activated, no originAddr is needed.
 */
-int activateRoute(uint32_t group, uint32_t originAddr, int upstrVif) {
+int activateRoute(uint32_t group, uint32_t originAddr, signed upstrVif) {
```
pali: `int` is by default signed and `signed` is just synonym for `int`. It is not common to write `signed` instead of `int`. So what is there reason for this change?

src/rttable.c
```diff
@@ -228,7 +240,7 @@ static void sendJoinLeaveUpstream(struct RouteTable* route, int join) {
 /**
 *   Clear all routes from routing table, and alerts Leaves upstream.
 */
-void clearAllRoutes(void) {
+void clearRoutes(struct IfDesc *IfDp) {
```
pali: unused function parameter?

pali:
> Also a logics table with few short actions, like this from createvifs:

That switch in `createVifs` is basically OK, optically it can be easier to read.

But such oneliners `for`/`if`
`for (struct IfDesc *Dp = TmpIfDescP.S; Dp < TmpIfDescP.E; Dp++) free(Dp->allowednets);`
`if (ptr->func) ptr->func(ptr->data);`
are not used in current igmpproxy code style. So they should be on two lines.

Generally it is not a good idea to bring own/new coding style to existing project to which are contributing more people. Doing it just result in having code very hard readable as every part of file would use different coding style (what people bring). Also reformating existing code in such project like igmpproxy is not a good idea, again if it would be done every contributor to his favourite coding style, then there would be only formatting changes.

src/config.c
```diff
@@ -100,6 +100,37 @@ struct Config *getCommonConfig(void) {
     return &commonConfig;
 }
 
+// Reloads the configuration file and removes interfaces which were removed from config.
+void reloadConfig() {
```
pali: igmpproxy is C not C++. Function with *no* arguments is declared in C as `void reloadConfig(void)`.

`void reloadConfig()` means function with unspecified arguments. It may be zero, ten, whatever.

src/callout.c
```diff
@@ -131,139 +88,68 @@ int timer_nextTimer(void) {
...
-    debugQueue();
-    return 0;
-}
-
 /**
  * debugging utility
  */
 static void debugQueue(void) {
-    struct timeOutQueue  *ptr;
+    struct timeOutQueue  *ptr; int i = 1;
```
pali: Put `i` on next own line. Thow are two different variables.

src/callout.c
```diff
@@ -131,139 +88,68 @@ int timer_nextTimer(void) {
...
 
     for (ptr = queue; ptr; ptr = ptr->next) {
```
pali: as `i` is used only in this `for`-loop it should be initialized in `for` and also incremented here:
`for (ptr = queue, i = 1; ptr; ptr = ptr->next, i++)`

src/config.c
```diff
@@ -100,6 +100,37 @@ struct Config *getCommonConfig(void) {
...
+
+    // Rebuild the interfaces config.
+    rebuildIfVc();
+
+    my_log(LOG_DEBUG, 0, "reloadConfig: Config Reloaded. OldConfPtr %x, NewConfPtr, %x", OldConfPtr, vifconf);
+    // Free all the old mallocd subnetlists and vifconf list.
+    while ( OldConfPtr ) {
+        TmpConfPtr=OldConfPtr->next; // Increment before free or pointers may be invalid.
+        struct SubnetList *TmpNetPtr;
+        for ( ; OldConfPtr->allowednets;  ) {
```
pali: `while`-loop?

src/config.c
```diff
@@ -100,6 +100,37 @@ struct Config *getCommonConfig(void) {
...
+            TmpNetPtr = OldConfPtr->allowednets->next;
+            free (OldConfPtr->allowednets);
+            OldConfPtr->allowednets = TmpNetPtr;
+        }
+        for ( ; OldConfPtr->allowedgroups; ) {
```
pali: `while`-loop?

src/config.c
```diff
@@ -260,6 +291,95 @@ void configureVifs(void) {
     }
 }
 
+/* create VIFs for all IP, non-loop interfaces.
+   When argument is not NULL rebuild the interface table.
+*/
+void createVifs(struct IfDescP *RebuildP) {
+    struct IfDesc *Dp, *oDp = NULL;
+    int    vifcount = 0, upsvifcount = 0, Ix = 0;
+    bool   join = 0;
```
pali: `false` and `true` are bool values. So either use `int join = 0;` or `bool join = false`.

src/ifvc.c
```diff
@@ -386,7 +260,7 @@ struct IfDesc *getIfByAddress( uint32_t ipaddr ) {
 *   the supplied IP adress. The IP must match a interfaces
 *   subnet, or any configured allowed subnet on a interface.
 */
-struct IfDesc *getIfByVifIndex( unsigned vifindex ) {
+struct IfDesc *getIfByVifIndex( signed vifindex ) {
```
pali: What does mean negative vifindex?

src/igmpproxy.c
```diff
@@ -289,17 +289,23 @@ void igmpProxyRun(void) {
...
 
-        /* aimwang: call rebuildIfVc */
-        if (config->rescanVif)
-            rebuildIfVc();
-
-        // Prepare timeout...
-        secs = timer_nextTimer();
-        if(secs == -1) {
-            timeout = NULL;
+        if (config->rescanVif) (void)0;
```
pali, line 290: This does nothing.

src/config.c
pali: Anyway, this function is not used in current pull request, so it it is not needed in this pull request and may be separated into other next pull requests (to have smaller changes).

src/igmpproxy.h
Uglymotha: ok, I will remove it from the pull request. It can probably be fixed by having types.h, netinet/in.h in os.h etc.

src/rttable.c
Uglymotha: As of yet. The paramater will be a pointer to (old) interface configuration during dynamic interface evaluation to remove the routes belonging to (upstream) interface if it has been removed from configuration. I think it is better to use this function for such purpose as removeRoute() does all sorts of checks, while when interface is being removed from config you'll want to indiscriminately remove all routes going over that interface.

src/igmpproxy.c
Uglymotha: not yet, i put it in there to prevent compiler warning about unused variable config (which I don't want to remove, as it really should be in there).. It will be used to initialize timer to rescan interfaces in system.

src/ifvc.c
Uglymotha: This was not my invention. In current version, array of upstream interfaces is initialized as          // init array to "not set"

        for ( Ix = 0; Ix < MAX_UPS_VIFS; Ix++)

        {

            upStreamIfIdx[Ix] = -1;

        }

It was however never used to compare old interfaces array to new interfaces, so that's why compiler now throws warning.

src/config.c
Uglymotha: when you think you have it all covered, you're still doing strange things sometimes. :)

src/config.c
Uglymotha: yes

src/igmpproxy.c
Uglymotha: removed, but it is still needed to prevent these errors
/usr/include/net/route.h:82:2: error: unknown type name 'u_long'; did you mean 'long'?
        u_long  rmx_locks;      /* Kernel must leave these values alone */
        ^
/usr/include/net/route.h:83:2: error: unknown type name 'u_long'; did you mean 'long'?
        u_long  rmx_mtu;        /* MTU for this path */
        ^

src/config.c
Uglymotha: fixed

src/igmpproxy.c
Uglymotha: Also, the option rescanvif is still being loaded by loadConfig(). So if somebody actually enables it, having the old if (config->rescanVif) rebuildIfVc() will cause bad things as it would in the current version, now it won't.

In newer version it will become:
// Set rescanvif or rescanconf timer.
        if (!config->rescanConf && config->rescanVif > 0 && timer_leftTimer(rescanvif_timer) == -1) {
            rescanvif_timer=timer_setTimer(config->rescanVif, (timer_f)rebuildIfVc, NULL);
        }
        if (config->rescanConf > 0 && timer_leftTimer(rescanconf_timer) == -1) {
            rescanconf_timer=timer_setTimer(config->rescanConf, (timer_f)reloadConfig, NULL);
        }

Uglymotha:
I think I have processed your comments and updated style of my code to be consistent. I also set up travis-ci on my own repo, so next pr's will be without warnings and cleaner (hopefully).

I still left in reloadconfig() function. It's not complex function and will make the next pr a little shorter. I think it will be a good trade-off between size and nr of prs needed to get in all my changes. I you really want to I'll remove it.

Waiting for travis to confirm build, after that's done I'm going to squash the commits.

I may also sync al the commits to my master branch and then open up a new clean pr when we agree on all the changes.

src/rttable.c
pali: Please introduce new variables only when you start really using it. Otherwise it is "dead code" (=unused). Also compiler would complain about declaring unused variable.

src/ifvc.c
pali: ok!

pali: anyway, you should use `int` type. `signed` is just alias for `int` and `int` is more common.

src/igmpproxy.c
pali: So it looks like you need to include some system specific header file prior to route.h...

Uglymotha: yes, I figured that out already, it has to do with order of including types.h netiniet/in.h etc.

src/rttable.c
Uglymotha: fixed this, removed the parameter from funtion

src/config.c
Uglymotha: fixed

src/callout.c
Uglymotha: fixed

src/config.c
Uglymotha: fixed

src/ifvc.c
Uglymotha: switched the declarations to int

src/callout.c
```diff
@@ -131,139 +90,72 @@ int timer_nextTimer(void) {
...
+    }
+    return -1;
 }
 
 /**
  * debugging utility
  */
 static void debugQueue(void) {
-    struct timeOutQueue  *ptr;
+    struct timeOutQueue  *ptr;
```
pali: useless whitespace change (added trailing space)

src/igmpproxy.h
```diff
@@ -155,7 +155,13 @@ struct IfDesc {
     unsigned int        robustness;
     unsigned char       threshold;   /* ttl limit */
     unsigned int        ratelimit;
-    unsigned int        index;
+    int                 index;
```
pali: Looking at those whole code and I think it does not make sense to change type of index from unsigned to signed. On lot of places in this pull request is changed index type from unsigned to signed integer, but all other places (e.g. getIfByVifIndex) can accept only valid interface index and therefore non-negative value -- which perfectly fits into unsigned type (which was used before).

Signed integer is used only because special value `-1` represents invalid value. But for this one case, special construct `(unsigned int)-1` is enough. We do not need to represent more other negative states.

src/igmpproxy.c
pali: So is `rescanvif` option with this change still working as before?

src/config.c
```diff
@@ -260,6 +293,115 @@ void configureVifs(void) {
...
+    for(Ix = 0; (Dp = getIfByIx(Ix)); Ix++) {
+        if (RebuildP == NULL) {
+            // Only add vif for valid interfaces on start-up.
+            if ((Dp->Flags & IFF_LOOPBACK) || (Dp->state != IF_STATE_DOWNSTREAM && Dp->state != IF_STATE_UPSTREAM)) {
+                continue;
+            }
+        } else {
+            /* Need rebuild, check if interface is new or already exists (check table below).
+                             old: disabled    new: disabled    -> do nothing
+                             old: disabled    new: downstream  -> addVIF(new), joinmcroutergroups
```
pali: This calls `joinMcRoutersGroup` function which uses `setsockopt`/`IP_ADD_MEMBERSHIP` API to join mutlicast groups 224.0.0.2 (INADDR_ALLRTRS_GROUP) and 224.0.0.22 (INADDR_ALLIGMPV3_GROUP).

But it is doing it on downstream interface.

So what is purpose of this call? And why it is needed?

src/igmpproxy.h
Uglymotha: you are right, it's back to unsigned now

src/config.c
Uglymotha: This has been split off from initRoutetable, which is run upon startup to:
   // If this is a downstream vif, we should join the All routers group...

So if you have interfaces transitioning from disabled to downstream, these groups need to be joined.

src/igmpproxy.c
Uglymotha: Well yes and no. Currently it is not working correctly. There are several issues. First it is not building a new list of interfaces to compare with old and acti accordingly. It is only making interfaces hidden, not going through the process of removing routes, sending leaves etc for upstream interfaces. It is not doing reevalution of groups joined when downstream interface transitions to disabled. The only thing currently done is:
leaveMcGroup( getMcGroupSock(), Dp, allrouters_group ); delvif
Second when interface are really dynamic, ie also constantly changing names like ppp1.....xxxx the function will keep creating new IfDesc ad infinitum and never do any cleaning, which will cause memory leaks. It's not good habit to keep references to stuff that has really disappeared.
Third, it is called on every event loop iteration, so effectively every event. Rebuilding the interfaces takes considerable time compared to handling a request. So this may lead to unwanted delays, if on a larger network there are many requests to be handled.

So instead of not working correctly, it is now just not working. It will be when the rest of my changes come through.

src/config.c
pali: But why it is needed to be joined? For me it does not make sense to join these two particular multicast groups on downstream interface.

Downstream interface just needs to listen to IGMP packets (not whole multicast traffic sent to 224.0.0.2 / 224.0.0.22) and this is achieved by calling `MRT_INIT` on `AF_INET`/`SOCK_RAW`/`IPPROTO_IGMP` socket. This socket receives all IGMP packets (with and also without Router Alter bit).

Morover on downstream interface igmpproxy needs to listen for igmppackets not only sent to 224.0.0.2 or 224.0.0.22 multicast groups, but needs to listen for **all** igmp packet sent to any multicast group. But above `MRT_INIT` handles it.

So for me joining just two particular groups does not make sense (it it not enough) and it is not needed for igmp packets as there is already `MRT_INIT`.

Uglymotha: IGMP does indeed not require those groups to be joined. This is for consistency in behaviour. According to the documentation IP_ADD_MEMBERSHIP should be set per interface, also delVif() is accounting for leaving these groups. That may lead to situation where only 1 interface is downstream and this being removed and replaced by another the groups will no longer be joined if they are not joined on the new interface. Hence I opted to implement this behaviour. IGMP wise this may be removed completely, as there is no requirement, nor need for the proxy to join.

pali: I think new `joinMcRoutersGroup` function can be removed as I do not see reason for it. Calling it from this code is not needed and when changing this part of code it is better to "remove old unneeded code" as trying to reimplement it.

Uglymotha: Ok I removed the functionality to join mcrouters groups completely. Because initRoutetable() was left with only initializing *routetable to NULL I moved initialization to global declration, removed initRoutetable() and use clearRoutes() in igmpproxyinit() for good order.

pali:
Hello! I see that you have added to this pull request new features like `rescanconf` or `loglevel` keys for config file. It is really intended for this PR about timings?

Uglymotha:
oh crap, I accidently did a PR merge on my own repo in the wrong direction. Sorry, I'm still learning to work with git.

Let's see if I can fix this. May be necessary to open yet another one.

pali:
You do not need to open a new pull requst. You can update any one. Just look for `git reset --hard` command and `git push --force` command. Second one will allow you to push any changes to github.

Uglymotha:
Awesome, the incorrect merge has ben reverted. PR is back in order again. I think we should be close on agreeing the changes.

I will add some lines of code the the createVifs() function as in the meantime I have worked out the last part of dynamic interface handling. If downstream if transitions to disabled or upstream the routes are now updated to clear the vif bit for that interface, zero downstream hosts, set to last member state and start the gsq process. Part of this is done in createvifs, so adding that to the code for this PR.

Uglymotha:
Over the pas week or so I have finished my rewrite. I also added black and whitelist functionality and the logic to reevaluate these on config reload. I also took the time to make everything unified in style and removed a lot of unused code. If you're interested take a lok at the rewrite branch in my personal repo.

The good thing is that we can now start changing everything from the inside out. So for cleanliness I will be opening another new smaller PR.

Issue #68, opened 2020-05-12
Compilation Errors on FreeBSD


Uglymotha:
On FreeBSD igmpproxy compilation throws errors like:
/usr/include/net/route.h:82:2: error: unknown type name 'u_long'; did you mean 'long'?
/usr/include/net/route.h:63:18: error: field has incomplete type 'struct sockaddr'
/usr/include/netinet/ip_mroute.h:98:17: error: field has incomplete type 'struct in_addr'

The reason for this is os-freebsd.h is missing several includes, as well as an early or incorrect define of __POSIX_C_SOURCE (which is done in igmpproxy.c before including igmpproxy.h. The latter is causing cdefs.h, included from types.h included from params.h to set BSD_VISIBLE to 0:

#elif defined(_C11_SOURCE)      /* Localism to specify strict C11 env. */
#define __POSIX_VISIBLE         0
#define __XSI_VISIBLE           0
#define __BSD_VISIBLE           0
#define __ISO_C_VISIBLE         2011
#define __EXT1_VISIBLE          0

In turn this causes types.h not providing the typedefs for u_long etc. The missing include of sys/param.h will also invalidate the evaluation of __FreeBSD_version because the file does not seem to be included from any of the other includes is os-freebsd.h (tested with cc-H os-freebsd.h).

I fixed this issue in os-freebsd.h by adding to top of file, will open a PR for this:
#undef _POSIX_C_SOURCE
#include <sys/param.h>
#define _POSIX_C_SOURCE 200112L

#include <sys/socket.h>
#include <netinet/in.h>

The missing include of sys/param.h will also invalidate the evaluation of

PR #69, opened 2020-05-12
Free bsd patch

Uglymotha:
fixes issue #68

### DIFF
--- a/src/os-freebsd.h
+++ b/src/os-freebsd.h
@@ -1,3 +1,7 @@
+#define __BSD_VISIBLE 1
+#include <sys/param.h>
+#include <sys/socket.h>
+#include <netinet/in.h>
 #include <net/route.h>
 #include <netinet/in_systm.h>
 #include <netinet/ip_mroute.h>


src/os-freebsd.h
```diff
@@ -1,3 +1,11 @@
+#define _POSIX_C_SOURCE_TMP _POSIX_C_SOURCE
+#undef _POSIX_C_SOURCE
+#include <sys/param.h>
+#define _POSIX_C_SOURCE _POSIX_C_SOURCE_TMP
+#undef _POSIX_C_SOURCE_TMP
```
pali: Hello! This code is incorrect and dos not work.

You cannot define old value of `_POSIX_C_SOURCE` which was prior to `#undef _POSIX_C_SOURCE` line. C macros are evaluated at the time of usage. So above code after fifth line would result to having `_POSIX_C_SOURCE` defined to string `_POSIX_C_SOURCE_TMP`. It is because `_POSIX_C_SOURCE_TMP` is undefined after fifth line. You can check/verify it e.g. by calling C preprocessor only:

Content of `file.c`:
```c
#define _POSIX_C_SOURCE "OLD_STRING_VALUE"

#define _POSIX_C_SOURCE_TMP _POSIX_C_SOURCE
#undef _POSIX_C_SOURCE

#define _POSIX_C_SOURCE _POSIX_C_SOURCE_TMP
#undef _POSIX_C_SOURCE_TMP

int main() { printf("%s", _POSIX_C_SOURCE); }
```
And call:
```
$ cpp test.c 
# 1 "test"
# 1 "<built-in>"
# 1 "<command-line>"
# 31 "<command-line>"
# 1 "/usr/include/stdc-predef.h" 1 3 4
# 32 "<command-line>" 2
# 1 "test"
# 9 "test"
int main() { printf("%s", _POSIX_C_SOURCE_TMP); }
```
It is not `"OLD_STRING_VALUE"` anymore.

pali (changes requested):

Uglymotha:
Oh crap you are right of course. 

Delved a little deeper. If _POSIX_C_SOURCE is only needed for clock_gettime from time.h and getopt from unistd.h, it may be undefined completely.
cdefs.h will set __POSIX_VISIBLE and __BSD_VISIBLE if neither _POSIX_C_SOURCE or _ANSI_SOURCE is defined and all will work as expected:
#else                           /* Default environment: show everything. */
#define __POSIX_VISIBLE         200809
#define __XSI_VISIBLE           700
#define __BSD_VISIBLE           1
#define __ISO_C_VISIBLE         2011
#define __EXT1_VISIBLE          1

pali:
Anyway, this looks like a bug in freebsd header files. Could you fire a bug to freebsd project? I think that freebsd developers should be informed about it and maybe also they come up with better workaround until they fix their system include files. But of course some workaround in igmpproxy for freebsd is needed...

Uglymotha:
It's not a bug, it's a feature. On FreeBSD when POSIX environment is requested, BSD specific features will not be made visible. According to FreeBSD community it is actually Linux that mishandles this. More eloborated upon in this thread:
https://lists.freebsd.org/pipermail/freebsd-standards/2004-March/000474.html

FreeBSD is so kind as to provide a default environment with all BSD and POSIX functionality available. I do not see any reason not to use this, since igmpproxy only needs clock_gettime and getopt, and they are available in the right format in the default environment.

pali:
For me it looks stupid if defining _POSIX_C_SOURCE cause that you are not able to include route.h file because it throws error `/usr/include/net/route.h:82:2: error: unknown type name 'u_long'; did you mean 'long'?`...

Anyway, I do not have freebsd box for testing this change. But if it is enough to workaround/fix that compilation issue, I'm fine to merge this pull request.

pali:
Ok, I looked at this problem and this is definitely bug in freebsd header files.

Their _sys/types.h_ contains following code:
```c
#if __BSD_VISIBLE
typedef	unsigned char	u_char;
typedef	unsigned short	u_short;
typedef	unsigned int	u_int;
typedef	unsigned long	u_long;
...
```
So `u_long` type is visible only when `__BSD_VISIBLE` is defined.

And their _net/route.h_ contains following code:
```c
struct rt_metrics {
	u_long	rmx_locks;	/* Kernel must leave these values alone */
	u_long	rmx_mtu;	/* MTU for this path */
...
```
So they define `rt_metrics` structure with `u_long` type always, even when `__BSD_VISIBLE` is not defined. Which leads to incorrect definitions and broken header inclusion.

Freebsd should fix this issue by e.g. changing `u_long` type in _net/route.h_ file to standard C `unsigned long` type. Or they should add `#ifdef __BSD_VISIBLE` guard around `struct rt_metrics {` definition.

So I think correct workaround until above freebsd issue is fixed, to add `#define __BSD_VISIBLE 1` into _src/os-freebsd.h_ file.

pali:
Ok, so if it is working, please squash commits into one commit with proper explaining commit message and I will merge it into master branch.

Uglymotha:
Squashed the commits, should be ready to pull now.

pali:
And can you put any information into commit message? It lacks reason and what is fixing...

Plus I see that you have changed fix without any explanation. So I really do not know what happened and why such change...

Anyway, I would suggest some nice documents to read, how to write commit message. It helps other people to not only review code but also in future to understand what patch/change does and why was needed...
https://www.kernel.org/doc/html/latest/process/submitting-patches.html#describe-your-changes
https://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines#New_Development

Uglymotha:
Commit message added with proper description. Fixup was changing #undef _POSIX_C_SOURCE to #define __BSD_VISIBLE 1

src/os-freebsd.h
```diff
@@ -1,3 +1,7 @@
+#undef _POSIX_C_SOURCE
+#include <sys/param.h>
+#include <sys/socket.h>
+#include <netinet/in.h>
```
pali: In commit message you are saying that you are defining `__BSD_VISIBLE` macro, but it is not defined in this pull request.

Also in commit message you have extra lines author/committer/date/changed files which are always present in commit, see it via command `git show --format=fuller --stat`

pali (changes requested):

src/os-freebsd.h
Uglymotha: Fixed, it should be correct now

pali: Seems you put some extra duplicate lines `Define __BSD_VISIBLE` and `define __BSD_VISIBLE instead of #undef _POSIX_C_SOURCE` into commit message.

Uglymotha: fixed

pali:
Ok! Please rebase it on top of igmpproxy master branch and I will merge it.

Uglymotha:
Do I really have to rebase it on my master branch? Because there I have the completely rewritten version. 

I also have a branch that is in sync with you master branch, can I rebase it on that?

pali:
Yes you can without any problem. Nonconflicting change can be rebased without any problem. It just make git history linear, like you can see in igmpproxy master branch.

Uglymotha:
Hi I have already moved my rewrite into a different branch and restored my master branch to be in sync with yours. My freebsd patch has been placed on top of that and commit history is correct there now.

So I will be opening another new PR for this and also for the other I have, so we can keep master branches synced from now on.

pali:
Ouuuu, really please use one pull request for one change. And not multiple pull requests for one change. Also not one pull request for multiple changes. It just makes mess and I need to sped more time for review. Also if you create a new pull request for existing change (and close old one) you just make more harder to track where was done previous review of code and makes reviewer life more harder...

PR #70, opened 2020-05-13
IGMPv3 SSM support

zocker007:
added support for IGMPv3 SSM. Filter mode INCLUDE should be fully suported, even with multiple upstream interfaces and several downstream hosts requesting different or the same sources for one group. BLOCK requests are ignored when the requesting host is not the last for this group.
Group-and-Source speific queries are not yet implemented.
Filter mode EXCLUDE is only supported with empty source list. Exclude lists from downstream hosts are Ignored. Also block requests are ignored in EXCLUDE mode
Change in filter mode for a group is only supported from INCLUDE to EXCLUDE when an additional host requests the specific group in EXCLUDE. Once in EXCLUDE the filter mode can't change when the host which requested EXCLUDE mode leaves the group.

Tested on OPNSense 20.1.6 cascaded behind FritzBox with several IGMPv3 and IGMPv2 downstream hosts with and without SSM (UPNP devices and German Telekom MagentaTV MR401)

included timeout fix from #59


pali:
Hello! Thank you for working on this. I have there already open pull request for rewiring timing https://github.com/pali/igmpproxy/pull/66 which I think will be in the conflict with your pull request.

It would be great to cooperate with @Uglymotha to resolve possible merge conflicts and code duplication.

Also this pull request is huge and would take for me too much time to review it. I see that in this pull request you have more different changes, e.g. IGMPv3 and filter support.

Could you please split different features, bug fixes, etc... into more separate pull requests? It would really help me if I could review and merge small changes as one big pull reuest.

zocker007:
Thanks for your reply.

Ok I try to split this PR into several parts.

Uglymotha:
Great stuff. Have looked through your PR is fair detail and it definitely looks like it's perfectly mergable with what I'm doing. Can you open a PR on my repo master branch too?

I am currently in the process of a rewrite of core functionality for configuration and dynamic interface handling. That's pretty much finished in my master branch. You can remove the fix for issue #59 as the whole main event loop and timing / callout queue has been rewritten to something much simpler and better.

Uglymotha:
Never mind, I have opened the PR myself. New to git, still discovering all the ins & outs. :)

src/os-linux.h
```diff
@@ -2,14 +2,16 @@
 #define _GNU_SOURCE
 #include <netinet/in.h>
 #include <netinet/ip.h>
-#include <netinet/igmp.h>
```
pali, line 5: What is the reason that you do not want to use de-facto standard netinet/igmp.h header file?

pali:
Hello! As I asked in https://github.com/pali/igmpproxy/pull/70#issuecomment-628235796 could you split changes into smaller and separate pull requests? Such big change is hard to review & merge.

zocker007:
Ok, sry for late reply. I will close this and split it up.

TheJulianJES:
Thank you all for doing this!

PR #71, opened 2020-05-20
Free BSD Compilation Patch

Uglymotha:
Fixes compilation erros on FreeBSD due to missing includes in os-freebsd.h
This was causing errors about undefined structs and types in missing includes.
Also defines __BSD_VISIBLE due to FreeBSD not providing u_int, u_long etc
in a default _POSIX_C_SOURCE environment.
    define  __BSD_VISIBLE 1
    include <sys/param.h>
    include <sys/socket.h>
    include <netinet/in.h>

### DIFF
--- a/src/os-freebsd.h
+++ b/src/os-freebsd.h
@@ -1,3 +1,7 @@
+#define __BSD_VISIBLE 1
+#include <sys/param.h>
+#include <sys/socket.h>
+#include <netinet/in.h>
 #include <net/route.h>
 #include <netinet/in_systm.h>
 #include <netinet/ip_mroute.h>


pali:
Issue `11` in commit message is not related to this pull request. This looks like some glitch.

Also commit message is somehow damaged, it contains incomplete/inconsistent lines:
```
include <sys/param.h>
include <sys/socket.h>
<netinet/in.h>
```

Uglymotha:
Weir indeed I did not put that in there. Anyway both PR and commit message are now fixed. Commit also refers to correct issue.

pali:
You do not need (and also should not) copy patch/change into commit message.

Also to mark commit or pull request that is fixing some issue use appropriate format for it. And do not invent new (putting number on the first line) which moreover is not supported by github. You should put on the own line line keyword `Fixes: ` followed by a url to issue. See git log history how it was done in previous changes.

Uglymotha:
fixed, hopefully it's fully compliant now

pali:
Now you have introduced into commit message extra right round bracket `Patch)` and extra colon before dot `os-freebsd.h:.`. This looks like you have a broken editor which puts random characters at random parts in plain text :-)

Uglymotha:
It was probably my keyboard, I.ve been banging on it too much. :)

pali:
Ok, thank you, merged!

PR #76, opened 2020-08-19
igmpproxy.c:203 Line cause max timeout is 3 sec,so this line should b…

cd-yangling:
…e the same behavior. otherwise the igmp query time interval will be wrong. slow timer is not 125s or quick timer is no 31 sec


pali:
This pull request seems to be superseded by https://github.com/pali/igmpproxy/pull/72

PR #77, opened 2020-12-17
allow building on macOS

kambala-decapitator:
This patch fixes build errors on macOS. Unfortunately, program doesn't work (I haven't investigated why), error message is:

> MRT_INIT failed; Errno(42): Protocol not available

The main point was to be able to edit code on my main working machine directly and check that it builds rather than in a Linux VM.

File `src/macos-ip_mroute.h` was taken from https://opensource.apple.com/source/xnu/xnu-792/bsd/netinet/ip_mroute.h with only a minor modification: had to enclose in comment a few identifiers after `#endif`.

Tested on macOS 10.14 (Mojave) with Xcode 11.3.1.

### DIFF
--- a/src/macos-ip_mroute.h
+++ b/src/macos-ip_mroute.h
@@ -1,0 +1,293 @@
+/*
+ * Copyright (c) 2000 Apple Computer, Inc. All rights reserved.
+ *
+ * @APPLE_LICENSE_HEADER_START@
+ * 
+ * The contents of this file constitute Original Code as defined in and
+ * are subject to the Apple Public Source License Version 1.1 (the
+ * "License").  You may not use this file except in compliance with the
+ * License.  Please obtain a copy of the License at
+ * http://www.apple.com/publicsource and read it before using this file.
+ * 
+ * This Original Code and all software distributed under the License are
+ * distributed on an "AS IS" basis, WITHOUT WARRANTY OF ANY KIND, EITHER
+ * EXPRESS OR IMPLIED, AND APPLE HEREBY DISCLAIMS ALL SUCH WARRANTIES,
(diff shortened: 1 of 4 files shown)


src/macos-ip_mroute.h
```diff
@@ -0,0 +1,293 @@
+/*
+ * Copyright (c) 2000 Apple Computer, Inc. All rights reserved.
+ *
+ * @APPLE_LICENSE_HEADER_START@
+ * 
+ * The contents of this file constitute Original Code as defined in and
+ * are subject to the Apple Public Source License Version 1.1 (the
+ * "License").  You may not use this file except in compliance with the
+ * License.  Please obtain a copy of the License at
+ * http://www.apple.com/publicsource and read it before using this file.
```
pali, line 10: Please read license properly prior using it.

**You may not use this file except in compliance with the License.**

This Apple Public Source License Version 1.1 is not compatible with GPLv2 or later license (under which is igmpproxy) and therefore you cannot use this file in igmpproxy.

It means that including this change into igmpproxy is violation of license arguments of copyright holder of this file and therefore changes in this pull requests cannot be used or included into igmpproxy project.

Apple Public Source License Version 1.1 is moreover [non-free license](https://www.gnu.org/philosophy/historical-apsl.html) and cannot be used in any free software (not only in igmpproxy).

pali (changes requested):

src/macos-ip_mroute.h
kambala-decapitator: Thank you for the response!

Sorry, didn't really dig info about this license... Guess I'll have to close the PR.

PR #81, opened 2021-06-29
Chroot and drop privileges after startup

bket:
With this PR:

- The apparent root directory will be changed after startup, thus
  denying igmpproxy access to files and commands outside that
  environmental directory tree. Normally this directory is "/var/empty"
  but that can be over-ridden with a new option in igmpproxy.conf.

- igmpproxy will drop root privileges after startup by changing id to
  another user. Normally this username is "nobody" but this can be
  over-ridden as well.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -196,6 +196,31 @@ int loadConfig(char *configFile) {
             // Read next token...
             token = nextConfigToken();
             continue;
+        }
+        else if(strcmp("chroot", token)==0) {
+            // path is in next token
+            token = nextConfigToken();
+
+            if (snprintf(commonConfig.chroot, sizeof(commonConfig.chroot), "%s",
+              token) >= (int)sizeof(commonConfig.chroot))
+                my_log(LOG_ERR, 0, "Config: chroot is truncated");
+
+            my_log(LOG_DEBUG, 0, "Config: chroot set to %s",
+              commonConfig.chroot);
+            token = nextConfigToken();
+            continue;
+        }
+        else if(strcmp("user", token)==0) {
+            // username is in next token
+            token = nextConfigToken();
+
+            if (snprintf(commonConfig.user, sizeof(commonConfig.user), "%s",
+              token) >= (int)sizeof(commonConfig.user))
+                my_log(LOG_ERR, 0, "Config: user is truncated");
+
+            my_log(LOG_DEBUG, 0, "Config: user set to %s", commonConfig.user);
+            token = nextConfigToken();
+            continue;
         } else {
             // Unparsable token... Exit...
             closeConfigFile();
--- a/src/igmpproxy.c
+++ b/src/igmpproxy.c
@@ -78,8 +78,11 @@ int     upStreamIfIdx[MAX_UPS_VIFS];
 */
 int main( int ArgCn, char *ArgVc[] ) {
 
-    int c;
+    int c, devnull = -1;
     bool NotAsDaemon = false;
+
+    struct Config *config = NULL;
+    struct passwd *pw = NULL;
 
     srand(time(NULL) * getpid());
 
@@ -140,18 +143,46 @@ int main( int ArgCn, char *ArgVc[] ) {
             break;
         }
 
+        // Open /dev/null before chrooting.
+        if (!NotAsDaemon) {
+            devnull = open("/dev/null", O_RDWR);
+            if (devnull == -1)
+                my_log(LOG_ERR, 0, "unable to open /dev/null");
+        }
+
+        config = getCommonConfig();
+
+        if (config->user[0]) {
+            pw = getpwnam(config->user);
+            if (pw == NULL)
+                my_log(LOG_ERR, 0, "unknown user %s", config->user);
+        }
+
+        if (config->chroot[0])
+            if (chroot(config->chroot) != 0 || chdir("/") != 0)
+                my_log(LOG_ERR, 0, "unable to chroot to %s",
+                  config->chroot);
+
+        if (pw != NULL)
+            if (setgroups(1, &pw->pw_gid) != 0 ||
+              setresgid(pw->pw_gid, pw->pw_gid, pw->pw_gid) != 0 ||
+              setresuid(pw->pw_uid, pw->pw_uid, pw->pw_uid) != 0)
+                my_log(LOG_ERR, 0, "unable to drop privileges to %s",
+                  config->user);
+
         if ( !NotAsDaemon ) {
 
             // Only daemon goes past this line...
             if (fork()) exit(0);
 
             // Detach daemon from terminal
-            if ( close( 0 ) < 0 || close( 1 ) < 0 || close( 2 ) < 0
-                || open( "/dev/null", 0 ) != 0 || dup2( 0, 1 ) < 0 || dup2( 0, 2 ) < 0
-                || setpgid( 0, 0 ) < 0
-            ) {
+            if ( close( 0 ) < 0 || close( 1 ) < 0 || close( 2 ) < 0 ||
+              dup2( devnull, 0 ) < 0 || dup2( devnull, 1 ) < 0 ||
+              dup2( devnull, 2 ) < 0 || setpgid( 0, 0 ) < 0 ) {
                 my_log( LOG_ERR, errno, "failed to detach daemon" );
             }
+            if (devnull > 2)
+                close(devnull);
         }
 
         // Go to the main loop.
--- a/doc/igmpproxy.conf.5.in
+++ b/doc/igmpproxy.conf.5.in
@@ -41,7 +41,29 @@ aliases not in use should be configured as disabled.
 Any line in the configuration file starting with
 .B #
 is treated as a comment. Keywords and parameters can be distributed over many lines.
-The configuration file has two main keywords:
+The configuration file has four main keywords:
+
+.B chroot
+.I directory
+.RS
+Changes the apparent root directory to the given path after startup, thus
+denying
+.B igmpproxy
+access to files and commands outside that environmental directory tree.
+.RE
+
+
+.B user
+.I username
+.RS
+Specifies the userid to which
+.B igmpproxy
+will change after startup.
+.B igmpproxy
+must be started as root, but it will drop root privileges to the specified
+user.
+.RE
+
 
 .B quickleave
 .RS 
--- a/src/igmpproxy.h
+++ b/src/igmpproxy.h
@@ -49,6 +49,9 @@
 #include <fcntl.h>
 #include <stdbool.h>
 #include <time.h>
+#include <grp.h>
+#include <limits.h>
+#include <pwd.h>
 
 #include <sys/socket.h>
 #include <sys/un.h>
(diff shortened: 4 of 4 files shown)


pali:
This pull request enforce usage of chroot and uid change. I do not think this is a good idea to disallow non-chroot usage.

bket:
@pali, I changed this PR. No changes to default behavior, thus no enforcement of chroot and uid change. The later two options can be set independently.

What do you think?

pali:
Chroot option does not work together with daemon mode. Few lines below after doing chroot and chdir is call which opens /dev/null file. But this file is not present in empty chroot.

bket:
Yes, stupid of me, I'm making a mess of this PR:

I have another diff in my tree, which replaces https://github.com/pali/igmpproxy/blob/master/src/igmpproxy.c#L145-L154 by 
```
if (daemon(1, 0 ) == -1)
    my_log( LOG_ERR, errno, "failed to detach daemon" );
```

I run tested this on OpenBSD, but it should work on NetBSD, FreeBSD and DragonflyBSD as well. Linux should be good with glibc since 2.21. I'm not sure if this covers all supported platforms?

I will force push the additions.

pali:
But you have still there same issue. You cannot open /dev/null even via glibc's daemon function if you chroot into empty directory.

bket:
I have only run tested this PR on OpenBSD, which does what I expect. `igmpproxy` is chrooted and drops privileges after startup, and it runs as daemon. FWIW I used the approach of `mcast-proxy` as example ([mcast-proxy/mcast-proxy.c#L128-L148](https://github.com/Esdenera/mcast-proxy/blob/master/usr.sbin/mcast-proxy/mcast-proxy.c#L128-L148)).

I will do some tests on Linux...if time permits.

src/igmpproxy.c
```diff
@@ -140,20 +140,36 @@ int main( int ArgCn, char *ArgVc[] ) {
...
+            if (strlen(config->user) != 0)
+                if (setgroups(1, &pw->pw_gid) != 0 ||
+                  setresgid(pw->pw_gid, pw->pw_gid, pw->pw_gid) != 0 ||
+                  setresuid(pw->pw_uid, pw->pw_uid, pw->pw_uid) != 0)
+                    my_log(LOG_ERR, 0, "unable to drop privileges to %s",
+                      config->user);
         }
 
+        if (!NotAsDaemon)
+            if (daemon(1, 0) == -1)
```
pali: Let me explain it again: After executing chroot and chdir you have no access to original root filesystem. So trying to open /dev/null or any other file from original file system either directly via syscall `open` or indirectly via `daemon` function just fails.

So you cannot use use neither `open` nor `daemon` function after calling `chroot` && `chdir`.

pali (changes requested):

src/igmpproxy.c
pali: You have to open all required files, including `/dev/null` prior executing `chroot` syscall.

bket:
@pali, thank you for your patience. You are right. I will try to come up with a solution.

For reference purposes: glibc's `daemon()` seems to do things different than `daemon()` on OpenBSD. I needed to see the difference with my own eyes, and prepared a small demonstrator.
```
#include <pwd.h>
#include <unistd.h>
#include <stdio.h>
#include <stdlib.h>

int main() {
	struct passwd *pw;

	if (geteuid() != 0) {
		printf("must be root\n");
		return 1;
	}

	pw = getpwnam("nobody");
	if (pw == NULL) {
		printf("getpwnam failed\n");
		return 1;
	}

	if (chroot("/var/empty") != 0 || chdir("/") != 0 ||
	    setgroups(1, &pw->pw_gid) != 0 ||
	    setresgid(pw->pw_gid, pw->pw_gid, pw->pw_gid) != 0 ||
	    setresuid(pw->pw_uid, pw->pw_uid, pw->pw_uid) != 0) {
		printf("drop priv failed\n");
		return 1;
	}

	if (daemon(1, 0) == -1) {
		printf("daemon failed\n");
		return 1;
	}

	sleep(5);
}
```
On OpenBSD
```
$ doas ./a.out; echo $?; ps -U nobody 
0
  PID TT  STAT        TIME COMMAND
20076 ??  S        0:00.00 ./a.out
```
On Arch
```
sudo ./a.out; echo $?
daemon failed
1
```
On Arch, after creating /dev/null in the chroot using `mknod`
```
sudo ./a.out; echo $?
0
```

pali:
> For reference purposes: glibc's `daemon()` seems to do things different than `daemon()` on OpenBSD.

Which is a good argument to not use `daemon()` function in multiplatform code. And currently it is not used.

src/igmpproxy.c
```diff
@@ -140,14 +143,41 @@ int main( int ArgCn, char *ArgVc[] ) {
...
+
         if ( !NotAsDaemon ) {
 
             // Only daemon goes past this line...
             if (fork()) exit(0);
 
             // Detach daemon from terminal
             if ( close( 0 ) < 0 || close( 1 ) < 0 || close( 2 ) < 0
-                || open( "/dev/null", 0 ) != 0 || dup2( 0, 1 ) < 0 || dup2( 0, 2 ) < 0
+                || dup2( devnull, 1 ) < 0 || dup2( devnull, 2 ) < 0
```
pali: You forgot for stdin.

pali (changes requested):

src/igmpproxy.c
bket: /dev/null is now opened before chrooting. Run tested on Alpine (musl libc), Arch (glibc) and OpenBSD.

src/igmpproxy.c
bket: Done...

src/igmpproxy.c
```diff
@@ -140,16 +143,42 @@ int main( int ArgCn, char *ArgVc[] ) {
             break;
         }
 
+        // Open /dev/null before chrooting.
+        if (!NotAsDaemon) {
+            devnull = open(_PATH_DEVNULL, O_RDWR, 0);
```
pali: `_PATH_DEVNULL` does not have to be available, so stick with `"/dev/null"` as it was before your change.

src/igmpproxy.c
```diff
@@ -140,16 +143,42 @@ int main( int ArgCn, char *ArgVc[] ) {
...
             // Only daemon goes past this line...
             if (fork()) exit(0);
 
             // Detach daemon from terminal
             if ( close( 0 ) < 0 || close( 1 ) < 0 || close( 2 ) < 0
-                || open( "/dev/null", 0 ) != 0 || dup2( 0, 1 ) < 0 || dup2( 0, 2 ) < 0
-                || setpgid( 0, 0 ) < 0
-            ) {
+                || dup2( devnull, 0 ) < 0 || dup2( devnull, 1 ) < 0
+                || dup2( devnull, 2 ) < 0 || setpgid( 0, 0 ) < 0 ) {
```
pali: This your change introduces a new resource leak. `devnull` file descriptor is not used after these calls and stay open forever when running in daemon mode.

src/config.c
```diff
@@ -196,6 +196,24 @@ int loadConfig(char *configFile) {
             // Read next token...
             token = nextConfigToken();
             continue;
+        }
+        else if(strcmp("chroot", token)==0) {
+            // path is in next token
+            token = nextConfigToken();
+            snprintf(commonConfig.chroot, sizeof(commonConfig.chroot), "%s",
+              token);
```
pali: There is missing check for directory overflow / truncate. When directory path is too long it should not be silently truncated, instead some fatal error.

src/config.c
```diff
@@ -196,6 +196,24 @@ int loadConfig(char *configFile) {
...
+            my_log(LOG_DEBUG, 0, "Config: Chroot set to %s",
+              commonConfig.chroot);
+            token = nextConfigToken();
+            continue;
+        }
+        else if(strcmp("user", token)==0) {
+            // username is in next token
+            token = nextConfigToken();
+            snprintf(commonConfig.user, sizeof(commonConfig.user), "%s", token);
```
pali: Same for username overflow / truncatenation.

pali (changes requested):
You still have there security related issues. Please fix them and also rebase pull request on top of master branch which will fix pull request CI testing.

src/igmpproxy.c
bket: Done.

src/igmpproxy.c
bket: Good catch! Issue has been addressed.

src/config.c
bket: Agreed! Return value of snprintf() is checked.

src/config.c
bket: Also here, issue has been addressed by checking return value of snprintf().

src/igmpproxy.c
```diff
@@ -140,18 +143,46 @@ int main( int ArgCn, char *ArgVc[] ) {
...
+            if (chroot(config->chroot) != 0 || chdir("/") != 0)
+                my_log(LOG_ERR, 0, "unable to chroot to %s",
+                  config->chroot);
+
+        if (strlen(config->user) != 0 && pw != NULL)
+            if (setgroups(1, &pw->pw_gid) != 0 ||
+              setresgid(pw->pw_gid, pw->pw_gid, pw->pw_gid) != 0 ||
+              setresuid(pw->pw_uid, pw->pw_uid, pw->pw_uid) != 0)
+                my_log(LOG_ERR, 0, "unable to drop privileges to %s",
+                  config->user);
```
bket: @pali, most builds in the CI run successfully. There are however a couple of build jobs that fail because the compiler (gcc-4.4 -> gcc-4.8) thinks a variable may be used uninitialized.
```igmpproxy.c: In function ‘main’:
igmpproxy.c:167:30: error: ‘pw’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
             if (setgroups(1, &pw->pw_gid) != 0 ||
                              ^
```
I do not understand why the compiler thinks this as the struct `pw` is checked before usage. I'm not sure that the code is at fault, and I'm thinking of selectively disabling this warning using something like below. What do you think, any ideas?
```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
...
#pragma GCC diagnostic pop
```

pali: Because you forgot to initialize it, on line 85 you have just:
```c
struct passwd *pw;
```
Fix it properly instead of adding compiler warnings hacks.

Also on line 166 you are testing `config->user` member but you are not using it at all. So you should not test it. Just `pw` is enough.

And instead of counting length of string via `strlen(config->user)` you just need to know if is non-empty, which is idiomatically tested just by `config->user[0]`.

bket: Thanks!

src/igmpproxy.c
```diff
@@ -140,18 +143,46 @@ int main( int ArgCn, char *ArgVc[] ) {
             break;
         }
 
+        // Open /dev/null before chrooting.
+        if (!NotAsDaemon) {
+            devnull = open("/dev/null", O_RDWR, 0);
```
pali: Third argument to `open` should be supplied only when `O_CREAT` or `O_TMPFILE` was specified.

bket: Fixed, and all tests pass.

pali:
Thanks, merged.

Issue #54, opened 2019-10-20
filter local multicast 239.255.255.250


poige:
Why?

Is there anything wrong with that?

pali:
Reference to discussion about this problem: https://github.com/pali/igmpproxy/commit/b445adbcbfa4b716e673ed55bfe53afc2f26baa8
And some proposed solutions: https://github.com/pali/igmpproxy/commit/b445adbcbfa4b716e673ed55bfe53afc2f26baa8#commitcomment-36374166

poige:
I agree that it's better to have it configurable, rather than built-in.

bb2020:
For those who are looking for a temporary solution: I install an obsolete version of `igmpproxy` on Openwrt (24kc device) to receive local SSDP. It is compatible with latest OS.

`opkg install https://downloads.openwrt.org/releases/17.01.0/packages/mips_24kc/base/
igmpproxy_0.1-10_mips_24kc.ipk`

pali:
Reopening, this issue is not fixed, I just linked discussion about this issue. If you would like to work on solution for this problem, feel free to propose patches / pull requests.

knightjoel:
Hi @pali, The patch below implements a new config option `allowssdp` which short circuits the check for the SSDP group. With this patch, I'm able to proxy reports for 239.255.255.250 between interfaces. If you think this is on the right track, I will fix up igmpproxy.conf(5) to mention this new config option.


```
--- src/config.c.orig   Sun Jan 10 14:10:07 2021
+++ src/config.c        Sun Jan 10 14:13:27 2021
@@ -196,6 +196,13 @@
             // Read next token...
             token = nextConfigToken();
             continue;
+        }
+        else if (strcmp("allowssdp", token)==0) {
+            my_log(LOG_DEBUG, 0, "Config: Will allow SSDP (239.255.255.250).");
+            commonConfig.allowSsdp = 1;
+
+            token = nextConfigToken();
+            continue;
         } else {
             // Unparsable token... Exit...
             closeConfigFile();
--- src/igmp.c.orig     Sun Jan 10 14:10:07 2021
+++ src/igmp.c  Sun Jan 10 14:13:27 2021
@@ -111,6 +111,8 @@
     struct igmpv3_grec *grec;
     int ipdatalen, iphdrlen, ngrec, nsrcs, i;

+    struct Config *config = getCommonConfig();
+
     if (recvlen < (int)sizeof(struct ip)) {
[...]

PR #82, opened 2021-07-07
Filter local multicast traffic (#54)

bket:
Implement new configuration option (filter), which enables filtering of traffic to a specified IP-address.


pali:
This configuration proposal is incorrect. This kind of filtering is for downstream interfaces, so it should be part of downstream interface section. Moreover there is already `whilelist` keyword and filtering code, so this filtering code should be extended to support e.g. `blacklist` keyword to allow configuring to accept all requests expects few (e.g. that SSDP)

pali:
Also one note: issue reference should not be in commit title, but rather on separate line as "Fixes: #num". This would make github automatically recognize it.

bket:
PR has been closed in favor of #83.

